PR #3, opened 2018-02-26
Allow for multiple IGVM tests work in parallel

TuxPowered42:


igvm/vm.py
```diff
@@ -119,6 +119,18 @@ def run(self, command, silent=False, with_sudo=True):
             else:
                 return super(VM, self).run(command, silent=silent)
 
+    def read_file(self, path):
+        """ Read a file from a running VM or a mounted image on HV
+        """
```
hasegeli: No space after `"""` and closing `"""` to the same line please.  See https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

igvm/vm.py
```diff
@@ -119,6 +119,18 @@ def run(self, command, silent=False, with_sudo=True):
...
+        with self.vm_host():
+            if self.mounted:
+                return self.hypervisor.read_file('{}/{}'.format(
+                    self.vm_path(''),
+                    path,
+                ))
+            else:
```
hasegeli: No need for the `else:`

tests/test_integration.py
```diff
@@ -69,163 +70,205 @@ def cmd(cmd, *args, **kwargs):
...
+    VM_route_net = Query(
+        {
+            'hostname': VM_NET,
+        },
+    ).get()['route_network']
+
+    # We can access HVs as objects but that does not mean we can compare them
+    # to any objects returned from igvm - those will be different objects,
+    # created from scratch from Serveradmin data.
+    HVs = [Hypervisor(h['hostname']) for h in Query({
```
hasegeli: Global constants are named upper case.  How about `HYPERVISORS`?

tests/test_integration.py
```diff
@@ -69,163 +70,205 @@ def cmd(cmd, *args, **kwargs):
...
+        'servertype': 'hypervisor',
+        'environment': 'testing',
+        'vlan_networks': VM_route_net,
+        'state': 'online',
+    })]
+
+    if len(HVs) < 2:
+        raise Exception('Not enough testing HVs found')
```
hasegeli: I don't like that we altering the global variables from here.  Why don't we put these to the beginning of the file like the other globals?

tests/test_integration.py
```diff
@@ -69,163 +70,205 @@ def cmd(cmd, *args, **kwargs):
...
-    except DatasetError:
-        pass
-    else:
-        vm_obj.delete()
-        vm_obj.commit()
+def setUpModule():
+    # Automatically find suitable HVs for tests.
+    # Terminate if this is impossible - we can't run tests without HVs.
+    global HVs
+    VM_route_net = Query(
```
hasegeli: `vm_route_net`

tests/test_integration.py
```diff
@@ -69,163 +70,205 @@ def cmd(cmd, *args, **kwargs):
...
     @classmethod
     def setUpClass(cls):
-        cls.hv1 = Hypervisor(HV1)
-        cls.hv2 = Hypervisor(HV2)
-        cls.vm = _create_vm()
+        cls.hvs = HVs
 
-        _clean_vm((cls.hv1, cls.hv2), cls.vm)
+    def setUp(self):
+        """ Initialize VM object before every test.
```
hasegeli: No space after `"""`

tests/test_integration.py
```diff
@@ -69,163 +70,205 @@ def cmd(cmd, *args, **kwargs):
...
+            'int:innogames:stable jessie',
+        ]
+        self.vm_obj['puppet_environment'] = None
+        self.vm_obj.commit()
 
-    @classmethod
-    def tearDownClass(cls):
-        _clean_vm((cls.hv1, cls.hv2), cls.vm)
+    def tearDown(self):
+        """ Clean up all HVs after every test. """
```
hasegeli: No space after `"""`

tests/test_integration.py
```diff
@@ -69,163 +70,205 @@ def cmd(cmd, *args, **kwargs):
...
 
-        self._check_vm(self.hv, self.vm)
+        for hv in self.hvs:
+            if (hv.server_obj['hostname'] == hv_name):
+                self.assertEqual(hv.vm_defined(vm), False)
+                hv.run('test ! -b /dev/xen-data/{}'.format(vm.fqdn))
 
 
 class BuildTest(IGVMTest):
+    """ Test many possible VM building scenarios """
```
hasegeli: No space around `"""`

tests/test_integration.py
```diff
@@ -445,225 +496,161 @@ def _get_vm():
...
         with self.assertRaises(IGVMError):
+            # Fake IP address is fine, this is a failing test.
             migratevm(
-                self.vm.server_obj['hostname'], HV2, offline=True, newip=IP2
+                self.vm_obj['hostname'], self.new_hv_name, offline=True,
+                newip='1.2.3.4',
             )
 
-    @unittest.skip("depends on automatic regeneration of ig.local domain")
+    @unittest.skip("Broken until we have new Serveradmin API deployed")
```
hasegeli: Let's try enabling this now.

TuxPowered42: Let's not enable it for now. I'm afraid we might loose hours trying to fix it. And since the test was always disabled, we don't make things worse.

tests/test_integration.py
TuxPowered42: Done.

igvm/vm.py
TuxPowered42: Fixed.

tests/test_integration.py
TuxPowered42: Fixed.

tests/test_integration.py
TuxPowered42: Fixed.

tests/test_integration.py
TuxPowered42: I don't see another way of exchanging information between setUpModule() and IGVMTest.setUpClass than a global variable.

hasegeli: How about putting these to the beginning of the file like the other globals and having no `setUpModule()`?

TuxPowered42: Having `setUpModule()` means that information about failure in preparation for tests is correctly reported to pytest. Unlike calling random code in random places with random `sys.exit()`.

hasegeli: Okay then we can keep it like this until we find a better solution.  Could you please at least move `setUpModule()` function to beginning of the file?

tests/test_integration.py
TuxPowered42: Renamed.

tests/test_integration.py
TuxPowered42: Done.

tests/test_integration.py
```diff
@@ -82,150 +109,166 @@ def _create_vm():
...
 
 
 class BuildTest(IGVMTest):
+    """Test many possible VM building scenarios."""
+
     def setUp(self):
-        self.hv = Hypervisor(HV1)
-        self.vm = _create_vm()
-        _clean_vm([self.hv], self.vm)
+        super(BuildTest, self).setUp()
```
HerrSpace: Why call it on super and not on self directly?

tests/test_integration.py
```diff
@@ -82,150 +109,166 @@ def _create_vm():
...
-
-    def test_simple(self):
-        buildvm(self.vm.server_obj['hostname'])
-
-        self.assertEqual(self.vm.hypervisor.fqdn, self.hv.fqdn)
-        self._check_vm(self.hv, self.vm)
-
-    def test_simple_stretch(self):
-        self.vm.server_obj.update({
+        super(BuildTest, self).tearDown()
```
HerrSpace: Same

tests/test_integration.py
```diff
@@ -235,155 +278,165 @@ def test_postboot(self):
...
 
         with self.assertRaises(IGVMError):
-            buildvm(self.vm.server_obj['hostname'])
+            buildvm(self.vm_obj['hostname'])
 
-        # Have we cleaned up?
-        self._check_absent(self.hv, self.vm)
+        self.check_vm_absent()
 
     def test_image_corruption(self):
```
HerrSpace: Maybe add a docstring, explaining that we expect igvm to redownload this image. I thought about this for a bit before realizing. Otherwise this test would require quite a bit of luck 🤔

tests/test_integration.py
```diff
@@ -235,155 +278,165 @@ def test_postboot(self):
...
+        # For every command test build a VM on the 1st HV
+        self.vm_obj['xen_host'] = self.hvs[0].server_obj['hostname']
+        self.vm_obj.commit()
+        buildvm(self.vm_obj['hostname'])
+        self.check_vm_present()
+        self.vm = VM(self.vm_obj)  # For contacting VM and HV over shell
+
     def tearDown(self):
-        _clean_vm(self.hv, self.vm.fqdn)
+        super(CommandTest, self).tearDown()
```
HerrSpace: Same

tests/test_integration.py
```diff
@@ -393,49 +446,47 @@ def _get_mem_vm():
...
+        self.vm_obj = self.get_vm_obj()
         self.assertEqual(_get_mem_hv(), 2048)
         self.vm.start()
         self.assertEqual(_get_mem_vm() - vm_mem, -1024)
 
     def test_vcpu_set(self):
-        buildvm(self.vm.server_obj['hostname'])
-
         def _get_hv():
             # Xen does not provide values when VM is powered off
```
HerrSpace: Outdated comment.

tests/test_integration.py
```diff
@@ -445,225 +496,161 @@ def _get_vm():
...
+        buildvm(self.vm_obj['hostname'])
+        # And is performed to the 2nd HV
+        # Of course apart from migrations to automatically selected HVs
+        self.new_hv_name = HYPERVISORS[1].server_obj['hostname']
 
     def tearDown(self):
-        # Make sure we leave with a good state
-        self._check_vm(self.hv1, self.vm)
-        self._check_absent(self.hv2, self.vm)
+        super(MigrationTest, self).tearDown()
```
HerrSpace: If I understand super correctly removing this method would not change any behavior.

igvm/settings.py
```diff
@@ -33,7 +33,7 @@
 # Mapping to determine the libvirt CPU model based on serveradmin hw_model
 KVM_HWMODEL_TO_CPUMODEL = {
     'Nehalem': ['Dell_M610', 'Dell_M710'],
-    'SandyBridge': ['Dell_M620', 'Dell_M630', 'Dell_R620'],
+    'SandyBridge': ['Dell_M620', 'Dell_M630', 'Dell_M640', 'Dell_R620'],
```
HerrSpace: This is not sandybridge though, right?

TuxPowered42: This is intended so that we can do live migrations between those systems.

tests/test_integration.py
TuxPowered42: No, this one is not the same. Although I removed methods which *only* call `super()`.

tests/test_integration.py
TuxPowered42: Same.

igvm/settings.py
HerrSpace: Hu, but we still also have M610s. Does that mean we can not migrate between those and the others either?

tests/test_integration.py
TuxPowered42: Cleaning XEN things was not scope of my work. We probably can even remove the code below, but let's do it in separate merge request.

tests/test_integration.py
TuxPowered42: Yes, this is the place which can be removed.

tests/test_integration.py
```diff
@@ -82,150 +109,166 @@ def _create_vm():
...
-
-class BalanceBuildTest(IGVMTest):
-    """ Same as BuildTest but with tearDown and setUp cleaning *both* HVs """
-
     @classmethod
     def setUpClass(cls):
-        cls.hv1 = Hypervisor(HV1)
-        cls.hv2 = Hypervisor(HV2)
-        cls.vm = _create_vm()
+        cls.hvs = HYPERVISORS
```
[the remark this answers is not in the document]
TuxPowered42: I rather hoped to never have any global variables at all, but since I found no better way to pass variable between `setUpModule()` and the rest of code, we go fully global now.

tests/test_integration.py
TuxPowered42: Done.

tests/test_integration.py
```diff
@@ -82,150 +109,166 @@ def _create_vm():
...
-        self.vm.server_obj['xen_host'] = None
-        self.vm.server_obj.commit()
+    def get_vm_obj(self):
+        vm_obj = Query(
+            {
+                'servertype': 'vm',
+                'hostname': VM_HOSTNAME,
+            },
+        )
+        return vm_obj.get()
```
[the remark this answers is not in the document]
TuxPowered42: Then we need real code in quite many places. Isn't this what functions are for, to not repeat real code?

tests/test_integration.py
HerrSpace: True, sorry for the noise.

tests/test_integration.py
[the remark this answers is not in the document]
HerrSpace: I just don't see the gain here, as the full query is already very brief. It's one less indirection I have to resolve in my head and it would probably still be less lines of code, but YMMV.

tests/test_integration.py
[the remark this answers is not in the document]
HerrSpace: Sure, not having globals is of course the best.

PR #14, opened 2018-03-13
Support Python 3

hasegeli:


hasegeli:
I made it depend on #13 to avoid rebase hassle.

igvm/commands.py
```diff
@@ -420,3 +536,60 @@ def vm_rename(vm_hostname, new_hostname, offline=False):
...
+
+def _get_vm(hostname, ignore_reserved=False):
+    """Get a server from Serveradmin by hostname to return VM object
+
+    The function is accepting hostnames in any length as long as it resolves
+    to a single server on Serveradmin.
+    """
+    conditions = [ExactMatch(hostname)]
+    if hostname.endswith('.ig.local'):
+        conditions.append(ExactMatch(hostname[:-len('.ig.local')]))
```
kofrezo: We should add a todo to remove this when we have all servers renamed properly

igvm/commands.py
```diff
@@ -145,10 +158,113 @@ def vm_build(vm_hostname, localimage=None, nopuppet=False, postboot=None,
...
+    if not was_running:
+        offline = True
+
+    if not offline and newip:
+        raise IGVMError('Online migration cannot change IP address.')
+
+    if not offline and runpuppet:
+        raise IGVMError('Online migration cannot run Puppet.')
+
+    if not runpuppet and newip:
```
kofrezo: Out of curiosity, whats the reason to run puppet but not change the IP on migration ?

igvm/migratevm.py
```diff
@@ -2,126 +2,9 @@
 
 Copyright (c) 2018, InnoGames GmbH
 """
+# XXX: This module is for backwards compatibility.
```
kofrezo: Where do we use this still ?

igvm/commands.py
hasegeli: This was actually to find the hosts without FQDN when the search is with FQDN.  We don't need this anymore.  So I am removing it, and putting the `TODO` to the place we append `.ig.local`.

igvm/commands.py
hasegeli: Maybe @TuxPowered42 knows?

igvm/migratevm.py
hasegeli: Any script that needs it.  This is the current place of the function.

igvm/commands.py
hasegeli: I prepared #18 for this.

igvm/commands.py
TuxPowered42: I don't know. It was always an option to run puppet after migration. I don't think there is any other use for it than to change IP address of a VM.

PR #22, opened 2018-03-22
Rip out sub-transactions

hasegeli:


igvm/drbd.py
```diff
@@ -171,75 +187,72 @@ def replicate_to_slave(self, transaction=None):
...
         self.hv.run('drbdadm create-md {}'.format(self.vm_name))
         self.hv.run('drbdadm up {}'.format(self.vm_name))
-        transaction.on_rollback(
-            'Bring DRBD device down', self.hv.run,
-            'drbdadm down {}'.format(self.vm_name)
-        )
-        self.hv.run('drbdadm wait-connect {}'.format(self.vm_name))
+        try:
+            self.hv.run('drbdadm wait-connect {}'.format(self.vm_name))
+        except Exception:
```
HerrSpace: I'm pretty sure run would raise a SystemExit on failure which is not a subclass of Exception but subclasses BaseException directly. This would therefore not work they way you expect, I think. See: https://github.com/fabric/fabric/issues/277#issuecomment-1861413

HerrSpace: Have we addressed this at all in igvm so far?

PR #23, opened 2018-03-23
drbd.py: Make StringIO import Python 2/3 compatible

kofrezo:


hasegeli:
See https://github.com/innogames/igvm/pull/22/commits

TuxPowered42:
@hasegeli , you admitted yesterday that your change will take long time to be finished. You also worked on Python 3 compatibility while simplifying transaction support. Let's instead take care of Python 3 compatibility separately and sooner.

hasegeli:
Here it is then: #24

PR #29, opened 2018-04-10
Fix image extraction or validation on build fails

kofrezo:
Commands to verify, download and extract image must have a exclusive
file lock and wait for each other. If not builds on the same hypervisor
might break each other e.g. when image gets downloaded, verfication
fails, images gets deleted and downloaded and while downloading another
build performs the same step.

I assume that 30 seconds is enough to download a image within a proper
network.

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -341,8 +341,10 @@ class Hypervisor(Host):
     def validate_image_checksum(self, image):
         """Compares the local image checksum against the checksum returned by
         foreman."""
+
+        image_file = '{}/{}'.format(IMAGE_PATH, image)
         local_hash = self.run(
-            'md5sum {}/{}'.format(IMAGE_PATH, image)
+            "flock -x -w 30 {0} -c 'md5sum {0}'".format(image_file)
         ).split()[0]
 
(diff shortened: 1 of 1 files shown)


igvm/hypervisor.py
```diff
@@ -316,8 +316,10 @@ def format_vm_storage(self, vm, transaction=None):
     def validate_image_checksum(self, image):
         """Compares the local image checksum against the checksum returned by
         foreman."""
+
+        image_file = '{}/{}'.format(IMAGE_PATH, image)
         local_hash = self.run(
-            'md5sum {}/{}'.format(IMAGE_PATH, image)
+            "flock -x -w 30 {} -c 'md5sum {}'".format(image_file, image_file)
```
HerrSpace: You can use {1} in both places and only pass image_file once.

kofrezo: {0} but I got the idea.

HerrSpace:
Retest this please

kofrezo:
Retest this please

TuxPowered42:
https://github.com/innogames/igvm/pull/43 seems a better solution.

PR #38, opened 2018-04-30
New url for foreman

mtkh:


igvm/settings.py
```diff
@@ -79,7 +79,7 @@
 IGVM_IMAGE_URL = environ.get(
     'IGVM_IMAGE_URL',
     # XXX: This default is deprecated, let's set it in puppet.
-    'http://aw-foreman.ig.local:8080/{image}'
+    'http://foreman-master.ndco.ig.local:8080/{image}'
```
hasegeli: This really doesn't belong to the open-source project.  Why don't we do what the comment says?  Please submit the merge request to `puppet/ig`.

HerrSpace: Using the ig::helper::environment in the igvm manifest would  be the way to go IMHO.

mtkh: Monday morning, first thing ;)

TuxPowered42: I don't think we should focus on this now when building new VMs is impossible. This can be rewritten at some later point and for now we will be just fine by only the URL change.

HerrSpace: Sure whatever goes in quicker. As these test seem broken again I expect that to be the puppet change though.

PR #46, opened 2018-05-21
Maintain object_id as a part of the libvirt domain and lv name

HerrSpace:
This MR changes the name of new libvirt domains and logical volumes from just the hostname to the object_id followed by the current hostname of the VMs serveradmin object . The object_id of a VM is immutable while the hostname is not. It's therefore much safer to correlate VMs by this ID. The domain and lv name of old VMs will be changed to the new version during an offline migration or when restarting the VM via igvm.
Once all VMs have been restarted, offline migrated or recreated we can remove the backward compatible hostname matching still maintained in this version.

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -87,7 +87,7 @@ class Hypervisor(Host):
         """
         with self.fabric_settings():
             self.lv_rename(
-                self.vm_lv_get(vm)['path'],
+                self.vm_lv_get(vm)['name'],
                 vm.uid_name
             )
 
@@ -480,10 +480,8 @@ class Hypervisor(Host):
                 .format(offline_transport)
             )
 
-        domain = self._get_domain(vm)
         if offline:
-            target_hypervisor.create_vm_storage(vm, vm.fqdn, transaction)
-
+            target_hypervisor.create_vm_storage(vm, vm.uid_name, transaction)
             if offline_transport == 'drbd':
                 self.start_drbd(vm, target_hypervisor)
                 try:
@@ -502,10 +500,10 @@ class Hypervisor(Host):
                     vm.shutdown(transaction)
                 with Transaction() as subtransaction:
                     nc_listener = target_hypervisor.netcat_to_device(
-                        self.vm_disk_path(vm.fqdn), subtransaction
+                        target_hypervisor.vm_lv_get(vm)['path'], subtransaction
                     )
                     self.device_to_netcat(
-                        self.vm_disk_path(domain.name()),
+                        self.vm_lv_get(vm)['path'],
                         vm.dataset_obj['disk_size_gib'] * 1024 ** 3,
                         nc_listener,
                         subtransaction,
@@ -513,7 +511,11 @@ class Hypervisor(Host):
 
             target_hypervisor.define_vm(vm, transaction)
         else:
-            target_hypervisor.create_vm_storage(vm, domain.name(), transaction)
+            target_hypervisor.create_vm_storage(
+                vm,
+                vm.hypervisor.vm_lv_get(vm)['name'],
+                transaction
+            )
             migrate_live(self, target_hypervisor, vm, self._get_domain(vm))
 
     def total_vm_memory(self):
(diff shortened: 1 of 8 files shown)


HerrSpace:
Restest this please

HerrSpace:
Restest this please

HerrSpace:
Retest this please

HerrSpace:
Retest this please

igvm/hypervisor.py
```diff
@@ -57,12 +57,27 @@ def __init__(self, dataset_obj):
         # We cannot store these in the VM object due to migrations.
         self._mount_path = {}
 
-    def vm_disk_path(self, name):
-        return '/dev/{}/{}'.format(VG_NAME, name)
+    def vm_lv_get(self, vm):
```
TuxPowered42: We name most of our functions as normal English sentences would go, e.g. "get_vlan_this", "format_vm_that", "download_those". Please stick to the rule.

igvm/hypervisor.py
```diff
@@ -57,12 +57,27 @@ def __init__(self, dataset_obj):
...
+        """Get a VMs logical volume information"""
```
TuxPowered42: "Gets VM's ..." or "Get ... of a VM".

igvm/hypervisor.py
```diff
@@ -57,12 +57,27 @@ def __init__(self, dataset_obj):
...
+        for lv in lvs:
+            if (
+                # Match the LV based on the object_id encoded within its name
+                vm.match_uid_name(lv['name']) or
+                # XXX: Deprecated matching for LVs w/o an uid_name
+                domain and lv['name'] == domain.name()
+            ):
+                return lv
+
+        raise HypervisorError(
```
TuxPowered42: I think StorageError would do better.

igvm/templates/domain.xml
```diff
@@ -1,12 +1,12 @@
 <domain type='kvm'>
-  <name>{{ fqdn }}</name>
+  <name>{{ name }}</name>
   <uuid>{{ props.uuid }}</uuid>
```
TuxPowered42: While we're at this.. why not use Serveradmin object ID as UUID for a VM?

tests/test_integration.py
```diff
@@ -148,17 +148,17 @@ def setUp(self):
...
         for hv in HYPERVISORS:
             hv.run(
                 'virsh destroy {vm}; '
-                'virsh undefine {vm}'
-                .format(vm=VM_HOSTNAME),
-                warn_only=True,
-            )
-            hv.run(
+                'virsh undefine {vm}; '
                 'umount /dev/xen-data/{vm}; '
```
TuxPowered42: While we're at it, `xen-data` should not be hardcoded but taken from settings.py

igvm/hypervisor.py
```diff
@@ -285,18 +300,20 @@ def vm_set_disk_size_gib(self, vm, new_size_gib):
...
             raise NotImplementedError('Cannot shrink the disk.')
-        domain = self._get_domain(vm)
         with self.fabric_settings():
-            self.lvresize(self.vm_disk_path(domain.name()), new_size_gib)
+            self.lv_resize(self.vm_lv_get(vm)['path'], new_size_gib)
 
         self._vm_set_disk_size_gib(vm, new_size_gib)
 
-    def create_vm_storage(self, vm, name, transaction=None):
+    def create_vm_storage(self, vm, lv_name, transaction=None):
```
TuxPowered42: Is it really necessary to give both VM and lv_name to create_vm_storage? Can't we move logic of choosing LV name inside this function?

igvm/templates/domain.xml
HerrSpace: > The content of the uuid element provides a globally unique identifier for the virtual machine. The format must be RFC 4122 compliant, eg 3e3fce45-4f53-4fa7-bb32-11f34168b82b.

HerrSpace: Object_id is not a valid uuid

igvm/hypervisor.py
HerrSpace: No, because we need the old lv name during online migrations and the new lv name during offline migrations and creations.

igvm/hypervisor.py
HerrSpace: Certainly not "Gets VM's ..." :P
I'll make it "Get logical volume information of a VM"

PR #47, opened 2018-05-23
Vlan tags rebased

TuxPowered42:


hasegeli:
Original one is rebased:

#21

PR #53, opened 2018-05-29
 Use the original user for the tests 

hasegeli:
We should be able to run the tests with our own users.


kofrezo:
Any update or progress here ?

TuxPowered42:
This was not resolved for over a year and IG employees don't really see it feasible.

PR #61, opened 2018-06-27
Increase read buffer, by default Fabric reads a byte at a time

TuxPowered42:


Felixoid:
It doesn't change anything
```
real	22m43.958s
user	14m47.912s
sys	7m17.044s
```

PR #65, opened 2018-07-17
Fix delete_vm() for legacy hosts with no uid prefix in their lv name

HerrSpace:
Patrick had a rebuild failing as get_volume_by_vm() matches first by
uid prefix and secondly by lv name == domain name. The latter does
not match as the domain is already deleted at this point in
delete_vm().

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -335,7 +335,6 @@ def vm_restart(vm_hostname, force=False, no_redefine=False):
             vm.shutdown()
 
         if not no_redefine:
-            vm.hypervisor.vm_lv_update_name(vm)
             vm.hypervisor.redefine_vm(vm)
 
         vm.start()
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -208,7 +208,7 @@ def vm_migrate(vm_hostname, hypervisor_hostname=None, newip=None,
             if vm.hypervisor.fqdn == hypervisor.fqdn:
                 raise IGVMError(
                     'Source and destination Hypervisor is the same!'
-                    )
+                )
         else:
             hypervisor = es.enter_context(_get_best_hypervisor(
                 vm,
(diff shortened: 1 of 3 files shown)


HerrSpace:
Fair enough I will see if it might even be save to delete the storage before undefining the VM tomorrow.
Until then I also added some changes that hopefully fix a related issue Serenus saw today.

HerrSpace:
I did not test `build --rebuild` as I don't have an old VM to do it with currently.
I tested this MR for:
- `vpu-set --offline` which worked perfectly
- `rename` which caused the following error unrelated to my changes:

```Disconnecting from 10.193.2.36... done.
Disconnecting from 10.1.254.6... done.
  File "./run.py", line 9, in <module>
    main()
  File "/home/patrick.meyer/igvm/igvm/cli.py", line 362, in main
    args.pop('func')(**args)
  File "/home/patrick.meyer/igvm/igvm/host.py", line 24, in decorator
    return fn(*args, **kwargs)
  File "/home/patrick.meyer/igvm/igvm/commands.py", line 541, in vm_rename
    vm.rename(new_hostname)
  File "/usr/lib/python3.4/contextlib.py", line 66, in __exit__
    next(self.gen)
  File "/home/patrick.meyer/igvm/igvm/commands.py", line 583, in _get_vm
    VM(vm_query(), hypervisor).release_lock()
  File "/home/patrick.meyer/igvm/igvm/commands.py", line 555, in vm_query
    }, VM_ATTRIBUTES).get()
  File "/usr/lib/python3/dist-packages/adminapi/dataset/__init__.py", line 111, in get
    raise DatasetError('get() requires exactly 1 matched object')
adminapi.dataset.DatasetError: get() requires exactly 1 matched object```

The renaming actually works perfectly. Only the unlocking of the VM fails as _get_vm is not aware of the changed hostname. These changes only make things better anyway.

HerrSpace:
Patrick also checked rebuilding and he says it works to <3

PR #68, opened 2018-07-20
Enforce smaller block size before DRBD migration starts

TuxPowered42:
Problem on VM:
```
XFS (vda): xfs_log_force: error 5 returned.
```

Problem on HV:
```
bio too big device drbd0 (16 > 8)
```

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -539,6 +539,23 @@ class Hypervisor(Host):
 
                 host_drbd = DRBD(self, vm, master_role=True)
                 peer_drbd = DRBD(target_hypervisor, vm)
+                if vm.hypervisor.vm_running(vm):
+                    vm_block_size = vm.get_block_size('/dev/vda')
+                    src_block_size = vm.hypervisor.get_block_size(
+                        vm.hypervisor.get_volume_by_vm(vm).path()
+                    )
+                    dst_block_size = target_hypervisor.get_block_size(
+                        target_hypervisor.get_volume_by_vm(vm).path()
+                    )
+                    log.debug(
+                        'Block sizes: VM {}, Source HV {}, Destination HV {}'
+                        .format(vm_block_size, src_block_size, dst_block_size)
+                    )
+                    vm.set_block_size('vda', min(
+                        vm_block_size,
+                        src_block_size,
+                        dst_block_size,
+                    ))
                 with host_drbd.start(peer_drbd), peer_drbd.start(host_drbd):
                     # XXX: Do we really need to wait for the both?
                     host_drbd.wait_for_sync()
--- a/igvm/host.py
+++ b/igvm/host.py
@@ -156,3 +156,29 @@ class Host(object):
     def release_lock(self):
         self.dataset_obj['igvm_locked'] = False
         self.dataset_obj.commit()
+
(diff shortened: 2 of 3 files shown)


igvm/hypervisor.py
```diff
@@ -536,6 +536,7 @@ def migrate_vm(
                         'DRBD migration is supported only between hypervisors '
                         ' using LVM storage!'
                     )
+                vm.safe_blocksize()
```
juergenthomann: I think we should find a different name for it. It currently could look like a getter (get me the safe blocksize for this vm). We already have some more methods starting with set_, so maybe use set_safe_blocksize()

igvm/vm.py
```diff
@@ -500,3 +500,15 @@ def unblock_autostart(self):
...
+    def safe_blocksize(self):
+        """ Reduce maximum number of KiB allowed for FS to request from
+            block layer
+
+        This is required for DRBD storage migrations. During disk migration
+        a DRBD will consist of devices on multiple hypervisors which might have
+        different size. It was observed that during high IO requests were
+        sometimes rejected by block layer causing filesystem to crash.
+        """
+        self.run('echo 64 > /sys/class/block/vda/queue/max_sectors_kb')
```
juergenthomann: Do checked for the lowest number? Maybe 128 is low enough for us.

igvm/hypervisor.py
TuxPowered42: Sure, I can fix that.

TuxPowered42:
retest this please

TuxPowered42:
retest this please

igvm/vm.py
```diff
@@ -500,3 +500,16 @@ def unblock_autostart(self):
...
+        """ Reduce maximum number of KiB allowed for FS to request from
+            block layer
+
+        This is required for DRBD storage migrations. During disk migration
+        a DRBD will consist of devices on multiple hypervisors which might have
+        different maximum block size. It was observed that during high IO
+        requests were sometimes rejected by block layer causing filesystem to
+        crash.
+        """
+        self.run('echo 64 > /sys/class/block/vda/queue/max_sectors_kb')
```
Woellchen: can we not query correct block size from the hvs taking part in the migration?

HerrSpace: From all the information I have I can't really follow how you even conclusively found out it's related to max_sector_kb. Maybe the write just fails at the moment we switch to the drbd device?

juergenthomann: It keeps failing, so it does not recover from an error until the server is killed.

igvm/hypervisor.py
```diff
@@ -536,6 +536,23 @@ def migrate_vm(
...
+                        vm.hypervisor.get_volume_by_vm(vm).path()
+                    )
+                    dst_block_size = target_hypervisor.get_block_size(
+                        target_hypervisor.get_volume_by_vm(vm).path()
+                    )
+                    log.debug(
+                        'Block sizes: VM {}, Source HV {}, Destination HV {}'
+                        .format(vm_block_size, src_block_size, dst_block_size)
+                    )
+                    vm.set_block_size('vda', min(
```
Woellchen: awesome

PR #70, opened 2018-07-24
Fix hypervisor not locked on auto selection

kofrezo:
This commit fixes that the hypervisor gets locked earlier than needed
and not when using autoselection.

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -163,9 +163,7 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
     with ExitStack() as es:
         vm = es.enter_context(_get_vm(vm_hostname))
 
-        if vm.hypervisor:
-            es.enter_context(_lock_hv(vm.hypervisor))
-        else:
+        if not vm.hypervisor:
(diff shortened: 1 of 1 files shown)


igvm/commands.py
```diff
@@ -180,6 +178,8 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
                 '"{}" is still running.'.format(vm.fqdn)
             )
 
+        es.enter_context(_lock_hv(vm.hypervisor))
```
TuxPowered42: _get_best_hyprevisor should return selected HV locked. With what you proposed there is a brief moment when HV is not locked. If _get_best_hypervisor does not return locked, it must be fixed.

kofrezo: The locking within this method is commented - for a reason I guess so if you are fine with adding it back in there that would be even better yes.

TuxPowered42: I think there is no locking in _get_best_hypervisor anyway. See https://github.com/innogames/igvm/pull/71

TuxPowered42: _get_best_hypervisor is a contextmanager so it would only release the lock after `yield` is finished.

PR #72, opened 2018-07-26
Avoid case, when LV was removed by lvremove

Felixoid:


TuxPowered42:
Why at all? Just don't use lvremove. It is not supported anymore especially since we have ZFS-based hypervisors. You should use virsh for storage operations.

PR #73, opened 2018-08-01
Restrict monitoring-worker on hypervisors to one

sponkydive:


igvm/settings.py
```diff
@@ -174,6 +174,8 @@
     # Don't migrate two masters database servers together
     OtherVMs(['game_world', 'function'], [0, 'db']),
     OtherVMs(['function'], ['master_db']),
+    # Don't migrate two monitoring worker to the same hypervisor
+    OtherVMs(['function'], ['monitoring-worker']),
```
hasegeli: We should try to generalize this somehow instead of hard-coding more `function`s in here.

PR #75, opened 2018-08-22
Commit VMs memory after point of no return

HerrSpace:
We have two sanity checks after increasing a VMs memory. However
should these checks fail we can not roll back as libvirt does not
support online decreasing the memory of a kvm domain. We therefore
have to commit the new memory value to serveradmin and can only warn
the user about the failed sanity checks.


TuxPowered42:
retest this please

PR #77, opened 2018-09-05
Fix killing of initial puppet runs in the KeyboardInterrupt handler

HerrSpace:
We have previously changed the puppet path in the run command but
failed to update it in the exception handler.


HerrSpace:
retest this please

PR #78, opened 2018-09-07
Format partitions for Stretch with XFS ftype support

kofrezo:


TuxPowered42:
retest this please

igvm/hypervisor.py
```diff
@@ -730,8 +731,11 @@ def umount_temp(self, device_or_path):
     def remove_temp(self, mount_path):
         self.run('rmdir {0}'.format(mount_path))
 
-    def format_storage(self, device):
-        self.run('mkfs.xfs -f {}'.format(device))
+    def format_storage(self, device, os):
+        if os == 'stretch':
```
TuxPowered42: This only verifies OS of VM but it should check HV too.

kofrezo:
We don't have it enabled on all VMs as showed to you @HerrSpace so I think it is still a good idea to enable it manually if it is aleady enabled because of CRC it does no harm.

HerrSpace:
@kofrezo I think we must find out on which HV the VM that didn't have it was build on. My guess is that HVs XFS just doesn't have it's feature and we rather must upgrade that. Maybe the VM was build on a Jessie HV?

igvm/hypervisor.py
```diff
@@ -730,8 +731,11 @@ def umount_temp(self, device_or_path):
     def remove_temp(self, mount_path):
         self.run('rmdir {0}'.format(mount_path))
 
-    def format_storage(self, device):
-        self.run('mkfs.xfs -f {}'.format(device))
+    def format_storage(self, device, os):
```
TuxPowered42: Wouldn't it be cleaner to give only True/False here for ftype parameter and determine it above?

TuxPowered42:
retest this please

TuxPowered42:
retest this pleae

TuxPowered42:
retest this please

PR #79, opened 2018-09-12
Don't shutdown VM on offline migrations, wait for operator to do it

TuxPowered42:


igvm/hypervisor.py
```diff
@@ -570,12 +571,21 @@ def migrate_vm(
                     vm.set_state('maintenance', transaction=transaction)
 
                     if vm.is_running():
-                        vm.shutdown(transaction)
+                        if no_shutdown:
+                            log.info('Please shut down the VM manually now')
+                            vm.wait_for_running(running=False, timeout=86400)
+                        else:
+                            vm.shutdown(transaction)
```
HerrSpace: Shouldn't we once more wait for wait_for_sync() right after the shutdown, especially now that the shutdown can be up to one day after the wait_for_sync above?

igvm/kvm.py
```diff
@@ -260,7 +247,6 @@ def migrate_live(source, destination, vm, domain):
         migrate_cmd.format(
             domain=domain.name(),
             destination=destination.fqdn,
-            timeout=timeout,
```
HerrSpace: We still have 1GbE HVs so what do I do on them if the VM never finishes now? Also how has this been a problem?

igvm/hypervisor.py
TuxPowered42: Not really. `wait_for_sync()` waits for initial "raid rebuild". After that is done, disk is in constant replication which is only torn down after VM is already stopped after `with` in line 567.

HerrSpace: Looking at the context manager start() and the called stop() in it's finally branch, I think there is also no logic to wait for the disk to be in sync after the shutdown. It seems like the stop branch just forcefully stops DRBD either way. Or am I missing something?

igvm/kvm.py
HerrSpace: Okay let's remove it and see if anybody actually complains about it being gone.

igvm/hypervisor.py
TuxPowered42: I am not 100% sure. I would have to investigate it and see what stop() really does. If there is a real bug, then it is in stop(), not here.

TuxPowered42: I was unable to find if `drbdadm down` ensures data consistency, but I changed protocol to C and re-enabled flushing, so any IO operation must be written down on both hosts before IO operation is finished.

PR #80, opened 2018-09-13
Connect to HV only when necessary

TuxPowered42:


TuxPowered42:
retest this please

tests/test_integration.py
```diff
@@ -154,20 +154,21 @@ def tearDown(self):
...
+            for vol_name in hv.get_storage_pool().listVolumes():
                 if vol_name == self.uid_name:
                     hv.run(
                         'mount | grep -q "/dev/{vg}/{vm}" && '
                         ' umount /dev/{vg}/{vm} || true;'
                         .format(vg=VG_NAME, vm=self.uid_name)
                     )
-                    hv.storage_pool.storageVolLookupByName(vol_name).delete()
+                    hv.get_storage_pool().\
+                        storageVolLookupByName(vol_name).delete()
```
HerrSpace: Escaping newlines is ugly. Split this in two lines, it makes for a nicer stack trace anyway.

igvm/hypervisor.py
```diff
@@ -55,12 +55,22 @@ def __init__(self, dataset_obj):
...
             )
 
+        self._mount_path = {}
+
+    def get_storage_pool(self):
         # Store per-VM path information
         # We cannot store these in the VM object due to migrations.
-        self._mount_path = {}
+        if hasattr(self, 'storage_pool'):
+            return self.storage_pool
```
HerrSpace: This smells like you want a property `storage_pool` with a getter doing what get_storage_pool() does.

If you want to keep get_storage_pool() you should set `storage_pool` to None in the constructor and check for `not None` instead of using hasattr for a cleaner interface.

Either way you should rename the class variable to _storage_pool as you don't want people to interact with it directly.

igvm/hypervisor.py
```diff
@@ -55,12 +55,22 @@ def __init__(self, dataset_obj):
...
         self.storage_pool = self.conn().storagePoolLookupByName(VG_NAME)
+        return self.storage_pool
+
+    def get_storage_type(self):
+        if hasattr(self, 'storage_type'):
+            return self.storage_type
```
HerrSpace: Same as above.

igvm/commands.py
```diff
@@ -611,20 +611,12 @@ def _get_hypervisor(hostname, ignore_reserved=False):
...
         'vlan_networks': vm.dataset_obj['route_network'],
         'state': Any(*hypervisor_states),
-    }, HYPERVISOR_ATTRIBUTES):
-        try:
-            hypervisors.append(Hypervisor(hv_obj))
-        except libvirtError:
-            log.warning(
-                'Could not connect to HV {}, skipping it'
-                .format(hv_obj['hostname'])
-            )
```
HerrSpace: I see why you are removing this here now. Shout out to @kofrezo for seeing this one first.

You still need to catch the libvirtError somewhere though. Maybe in self.conn()?

TuxPowered42: The place where connection might be initiated and fail is `hypervisor.check_vm()` at line 634. This basically return things to previous behaviour, which was correct.

kofrezo: Still there might be an libvirt error raised by the connection as @HerrSpace mentioned we better cover it. I think catching it at conn method is a good idea.

TuxPowered42: No, libvirtError is never raised by Hypervisor constructor anymore.

HerrSpace: Ah, indeed. Cool!

PR #83, opened 2018-09-23
Provide parameters to retire and unretire a VM

TuxPowered42:


igvm/cli.py
```diff
@@ -266,6 +266,11 @@ def parse_args():
         'vm_hostname',
         help='Hostname of the guest system',
     )
+    subparser.add_argument(
+        '--unretire',
+        action='store_true',
+        help='Unretire a VM, set it in maintenance mode',
+    )
```
kofrezo: I think it would be easier to just have a target state switch and default one would be maintenance

TuxPowered42: True, I'll implement it.

igvm/commands.py
```diff
@@ -315,6 +323,11 @@ def vm_stop(vm_hostname, force=False):
         else:
             vm.shutdown()
         log.info('"{}" is stopped.'.format(vm.fqdn))
+        if retire:
+            vm.dataset_obj['state'] = 'retired'
+            vm.dataset_obj.commit()
+            log.info('"{}" is retired.'.format(vm.fqdn))
+
```
HerrSpace: Setting the state before shutdown would make testtool take the VM out of potential LBs before it goes down. Can we move this?

igvm/commands.py
```diff
@@ -202,7 +202,7 @@ def vm_migrate(vm_hostname, hypervisor_hostname=None, newip=None,
         vm = es.enter_context(
             _get_vm(
                 vm_hostname, ignore_reserved=ignore_reserved,
-                ignore_retired=True,
+                allow_retired=True,
```
HerrSpace: `ignore_reserved` is talking about the hv and `allow_retired` about the vm. Maybe we can call them `ignore_reserved_

I propose renaming `ignore_reserved` into `allow_reserved_hv` and `allow_retired` into `allow_retired_vm`.

TuxPowered42: I think there is something very wrong with the code. `ignore_reserved` passed to `_get_vm()` has no effect on anything at all. I will just rip it out.

igvm/commands.py
TuxPowered42: Sure, I can do it.

TuxPowered42:
retest this please

TuxPowered42:
retest this please

igvm/commands.py
TuxPowered42: Ripped out.

igvm/commands.py
TuxPowered42: It is done.

igvm/cli.py
TuxPowered42: It is implemented.

PR #84, opened 2018-09-26
Add command to evacuate all VMs out of a hypervisor

kofrezo:
Add a command to evacuate all VMs from a hypervisor and set it to maintenance. Also supports migrating VMs offline or only specific VMs matching function attribute offline. Also adds dry mode.

### DIFF
--- a/igvm/cli.py
+++ b/igvm/cli.py
@@ -12,6 +12,7 @@ from fabric.network import disconnect_all
 
 from igvm.commands import (
     disk_set,
+    evacuate,
     host_info,
     mem_set,
     vcpu_set,
@@ -354,6 +355,26 @@ def parse_args():
         '--offline',
         action='store_true',
         help='Shutdown VM, if running',
+    )
+
+    subparser = subparsers.add_parser(
+        'evacuate',
+        description=evacuate.__doc__,
+    )
+    subparser.set_defaults(func=evacuate)
+    subparser.add_argument(
+        'hv_hostname',
+        help='Hostname of the hypervisor',
+    )
+    subparser.add_argument(
+        '--dry-run',
+        action='store_true',
+        help='Do not migrate but just print what would be done'
+    )
+    subparser.add_argument(
+        '--offline',
+        nargs='*',
+        help='Migrate VMs offline',
     )
 
     return vars(top_parser.parse_args())
--- a/igvm/cli.py
+++ b/igvm/cli.py
@@ -374,7 +374,7 @@ def parse_args():
     subparser.add_argument(
         '--offline',
         nargs='*',
-        help='Migrate VMs offline',
+        help='Migrate VMs matching the given serveradmin function offline',
     )
 
     return vars(top_parser.parse_args())
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -62,32 +62,32 @@ def evacuate(hv_hostname, offline=None, dry_run=False):
     a list of strings is passed only those matching will be migrate offline.
     """
 
-    hv = Query({'hostname': hv_hostname}, ['state'])
-    if len(hv) == 0:
-        print('No such hypervisor {}'.format(hv_hostname))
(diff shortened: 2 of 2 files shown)


igvm/commands.py
```diff
@@ -51,6 +51,45 @@ def _check_defined(vm, fail_hard=True):
...
+    a list of strings is passed only those matching will be migrate offline.
+    """
+
+    hv = Query({'hostname': hv_hostname}, ['state'])
+    if len(hv) == 0:
+        print('No such hypervisor {}'.format(hv_hostname))
+        exit(1)
+
+    if dry_run:
+        print('Setting {} to state maintenance'.format(hv_hostname))
```
TuxPowered42: This should be done only after HV is `igvm_locked`.

kofrezo: I adjusted it. Can you please check if I used the context manager for locking hv correctly ?

HerrSpace: I had a quick glance at it and think it should work. That said using an exit stack is not required if you only want to acquire one context manager. Therefore you can replace

```python3
    with ExitStack() as es:
        hv = es.enter_context(
            _get_hypervisor(hv_hostname, ignore_reserved=True)) 
```

with the following to the same end:
```python3
    with _get_hypervisor(hv_hostname, ignore_reserved=True) as hv:
        [...]
```

I wrote up this little piece about exit stacks and context managers after Emres presentation to be able look that stuff up again. Maybe it's helpful to you too :)
https://hackmd.io/ys74lBqQS5SuxIBMEHIH_Q#ExitStack

HerrSpace:
retest this please

kofrezo:
retest this please

igvm/commands.py
```diff
@@ -51,6 +51,42 @@ def _check_defined(vm, fail_hard=True):
...
+    Move all VMs out of a hypervisor and put it to state maintenance.
+
+    Offline can be passed without arguments or with a list strings matching
+    function attributes. If just passed all VMs will be migrated offline. If
+    a list of strings is passed only those matching will be migrate offline.
+    """
+
+    with _get_hypervisor(hv_hostname, ignore_reserved=True) as hv:
+        if dry_run:
+            log.info('Setting {} to state maintenance'.format(hv_hostname))
```
TuxPowered42: The message in dry-run should be the opposite, saying "not setting" or "I would set". Also there is no message at all when no dry-run happens. This relates to all messages below.

igvm/cli.py
```diff
@@ -356,6 +357,26 @@ def parse_args():
...
+    )
+    subparser.add_argument(
+        '--dry-run',
+        action='store_true',
+        help='Do not migrate but just print what would be done'
+    )
+    subparser.add_argument(
+        '--offline',
+        nargs='*',
+        help='Migrate VMs offline',
```
TuxPowered42: This help message is not really that helpful. First I wondered why it is `nargs=*` but only then I noticed it is compared against server functions. Yet the help message suggests it is a boolean.

igvm/commands.py
```diff
@@ -51,6 +51,42 @@ def _check_defined(vm, fail_hard=True):
...
+
+        for vm in hv.dataset_obj['vms']:
+            vm_function = vm['function']
+
+            if (
+                offline is not None and
+                (offline == [] or vm_function in offline)
+            ):
+                if dry_run:
+                    log.info('migrating {} offline'.format(vm['hostname']))
```
TuxPowered42: There are still more messages saying things are done when they are not.

igvm/commands.py
```diff
@@ -51,6 +51,42 @@ def _check_defined(vm, fail_hard=True):
...
+                else:
+                    vm_migrate(vm['hostname'], offline=True)
+            else:
+                if dry_run:
+                    log.info('migrating {} online'.format(vm['hostname']))
```
TuxPowered42: And here too.

PR #85, opened 2018-10-12
Prefer hypervisor with less I/O utilization

kofrezo:
Prefer to select a hypervisor with less I/O utilization as target for
migrations.


mtkh:
retest this please

PR #89, opened 2018-11-06
Allow deletion and stopping of vms in retired state

mtkh:


mtkh:
ok, I will update the MR accordingly afterwards

mtkh:
I have updated the parameters with new names

mtkh:
retest this please

igvm/commands.py
```diff
@@ -332,7 +332,7 @@ def vm_start(vm_hostname):
 @with_fabric_settings
 def vm_stop(vm_hostname, force=False):
     """Gracefully stop a VM"""
-    with _get_vm(vm_hostname) as vm:
+    with _get_vm(vm_hostname, allow_retired=True) as vm:
```
TuxPowered42: A VM which is retired must already be stopped. No live VM can be retired.

PR #94, opened 2018-11-26
Move IP address change out of migration

TuxPowered42:


igvm/cli.py
```diff
@@ -181,6 +177,23 @@ def parse_args():
             ' operator to shut down VM.'
         ),
     )
+
+    subparser = subparsers.add_parser(
+        'change-address',
```
kofrezo: change_address: underscore not dash

igvm/commands.py
```diff
@@ -183,6 +185,43 @@ def disk_set(vm_hostname, size):
...
+
+        if old_address == new_address:
+            raise ConfigError('New IP address is the same as the old one!')
+
+        vm_was_running = vm.is_running()
+
+        vm.dataset_obj['intern_ip'] = new_address
+        vm.dataset_obj.commit()
+
+        if vm_was_running:
```
kofrezo: Maybe it is a good idea to at least write in the comment that on IP change the VM gets rebootet.

igvm/commands.py
```diff
@@ -183,6 +185,43 @@ def disk_set(vm_hostname, size):
...
+    with _get_vm(vm_hostname) as vm:
+        old_address = vm.dataset_obj['intern_ip']
+        new_address = ip_address(new_address)
+
+        if old_address == new_address:
+            raise ConfigError('New IP address is the same as the old one!')
+
+        vm_was_running = vm.is_running()
+
+        vm.dataset_obj['intern_ip'] = new_address
```
kofrezo: How about IPv6 ?

tests/test_integration.py
```diff
@@ -88,6 +104,40 @@ def setUpModule():
 
     query.commit()
 
+    # Cancelled builds are forcefully killed by Jenkins. They did not have the
+    # opportunity to clean up so we forcibly destroy everything found on any HV
+    # which would interrupt our work in the current JENKINS_EXECUTOR.
+    for hv in HYPERVISORS:
+        hv.get_storage_pool().refresh()
+        for domain in hv.conn().listAllDomains():
+            if match(('^([0-9]+_)?' + VM_HOSTNAME_PATTERN + '$').format(
```
kofrezo: * You don't want to check for VM_HOSTNAME_PATTERN = 'igvm-{}-{}.test.ig.local' do you ? 
* How about parallel running builds ?

igvm/commands.py
TuxPowered42: I thought of something else: force this command to require `--offline`. That would make it consistent with other commands we have.

igvm/cli.py
kofrezo: True I was mixing things up with the positional argument.

igvm/commands.py
kofrezo: This is fine too.

PR #95, opened 2018-12-09
Use libvirt API for migration, not shell commands

TuxPowered42:


igvm/kvm.py
```diff
@@ -243,14 +243,70 @@ def migrate_live(source, destination, vm, domain):
...
-            domain=domain.name(),
-            destination=destination.fqdn,
-        ),
-        with_sudo=False,
+    migrate_flags = (
+        VIR_MIGRATE_LIVE |  # Do it live
+        VIR_MIGRATE_PERSIST_DEST |  # Define the VM on the new host
+        VIR_MIGRATE_CHANGE_PROTECTION |  # Protect source VM
+        VIR_MIGRATE_NON_SHARED_DISK |  # Copy non-shared storage
+        VIR_MIGRATE_AUTO_CONVERGE |  # Slow down VM if can't migrate memory
```
kofrezo: What is the effect here ? How does it "slow down" the VM ?

igvm/kvm.py
```diff
@@ -243,14 +243,70 @@ def migrate_live(source, destination, vm, domain):
...
+                log.info('Waiting for migration stats to show up')
+            time.sleep(1)
+    except KeyboardInterrupt:
+        log.warning('Cancelling migration')
+        domain.abortJob()
+
+    log.info('Awaiting migration to finish')
+    try:
+        future.result()
+    except Exception as e:
```
kofrezo: Do we have a more specific Exception we can catch ?

HerrSpace:
Looks good.

igvm/kvm.py
TuxPowered42: Also this is the exact copy of the old parameters. We might change them to post-copy migration for performance reasons later.

PR #98, opened 2019-01-24
Log hostname of the vm being migrated

mtkh:


mtkh:
rested this please

mtkh:
retest this please

igvm/kvm.py
```diff
@@ -268,7 +268,7 @@ def migrate_live(source, destination, vm, domain):
         (source.dataset_obj['os'], destination.dataset_obj['os'])
     )['flags']
 
-    log.info('Starting migration')
+    log.info('Starting migration of server: {}'.format(vm))
```
TuxPowered42: Why only only for live migration?

mtkh: good idea, its now logging for offline migration as well

PR #101, opened 2019-02-26
Fix host objects lock not released after crash

kofrezo:
We experienced that some objects such as hypervisors and VMs are not
porperly release their lock when igvm crashed or gets aborted by closing
the terminal or hitting ctrl-c many times in a row or the process gets
killed etc.

To get rid of this problem we should change igvm_locked to a number value
and work with timestamps so that we can check if a object has been locked
for too long as it is currently not possible to query serveradmin
history via API and this requires greater refactoring.

### DIFF
--- a/igvm/host.py
+++ b/igvm/host.py
@@ -4,6 +4,7 @@ Copyright (c) 2018 InnoGames GmbH
 """
 
 from io import BytesIO
+from datetime import datetime
 
 import fabric.api
 import fabric.state
@@ -138,13 +139,13 @@ class Host(object):
             )
 
     def acquire_lock(self, allow_fail=False):
-        if self.dataset_obj['igvm_locked']:
+        if self.dataset_obj['igvm_locked'] is not None:
             raise InvalidStateError(
                 'Server "{0}" is already being worked on by another igvm'
                 .format(self.dataset_obj['hostname'])
             )
 
-        self.dataset_obj['igvm_locked'] = True
+        self.dataset_obj['igvm_locked'] = int(datetime.utcnow().timestamp())
         try:
             self.dataset_obj.commit()
         except DatasetError:
@@ -154,7 +155,7 @@ class Host(object):
             )
 
     def release_lock(self):
-        self.dataset_obj['igvm_locked'] = False
+        self.dataset_obj['igvm_locked'] = None
         self.dataset_obj.commit()
 
     def get_block_size(self, device):


kofrezo:
This requires the serveradmin attribute igvm_locked to be changed to a number and rollout new igvm version everywhere it is used in sync.

igvm/host.py
```diff
@@ -154,7 +155,7 @@ def acquire_lock(self, allow_fail=False):
             )
 
     def release_lock(self):
-        self.dataset_obj['igvm_locked'] = False
+        self.dataset_obj['igvm_locked'] = None
```
TuxPowered42: Is this the same as delattr()?

igvm/host.py
```diff
@@ -138,13 +139,13 @@ def put(self, remote_path, local_path, mode='0644'):
...
     def acquire_lock(self, allow_fail=False):
-        if self.dataset_obj['igvm_locked']:
+        if bool(self.dataset_obj['igvm_locked']):
             raise InvalidStateError(
                 'Server "{0}" is already being worked on by another igvm'
                 .format(self.dataset_obj['hostname'])
             )
 
-        self.dataset_obj['igvm_locked'] = True
+        self.dataset_obj['igvm_locked'] = int(time())
```
TuxPowered42: Is this timestamp in users' local timezone? Can we enforce it is in UTC?

It would be way nicer if Serveradmin had a now() function, but I understand this is beyond scope of this ticket.

kofrezo: Indeed, I changed it to use UTC timestamp.

igvm/host.py
kofrezo: This is okay. Not set attributes are None so setting it to None is the correct way.

`>>> from adminapi.dataset import Query
>>> print(Query().new_object(servertype='vm'))
DatasetObject({'responsible_admin': MultiAttr(), 'no_monitoring': False, 'project_network': None, 'puppet_disabled': False, 'datacenter': None, 'floating_addresses': MultiAttr(), 'nagios_groups': MultiAttr(), 'servertype': 'vm', 'ssh_environment': None, 'monitoring_queue': None, 'iggop_locked': False, 'project_domain': None, 'disk_size_gib': 6, 'nagios_admins': MultiAttr({'admins'}), 'repositories': MultiAttr({'int:innogames:stable'}), 'backup_prio': 300, 'puppet_classes': MultiAttr(), 'project': None, 'acl': MultiAttr(), 'game_type': None, 'auto_update': True, 'allow_to': MultiAttr(), 'served_markets': MultiAttr(), 'backup_storage': None, 'intern_ip': None, 'provider_network': None, 'backup_canshrink': False, 'os': 'stretch', 'backup_window': MultiAttr({'2-5'}), 'monitoring_queues': MultiAttr(), 'backup_sum': 3, 'monitoring_admins': MultiAttr(), 'bladecenter_slot': MultiAttr(), 'mac': MultiAttr(), 'instance': None, 'object_id': None, 'ipv4_only': False, 'legal_personal_data': None, 'network_zones': MultiAttr(), 'team': None, 'ssh_groups': MultiAttr(), 'num_cpu': 4, 'memory': 2048, 'description': None, 'puppet_master': 'master-puppet.ig.local', 'security_confidentiality': None, 'backup_disabled': False, 'hypervisor_standby': None, 'loadbalancer': MultiAttr(), 'environment': 'production', 'monitoring_checks': MultiAttr({'ping'}), 'os_product_license': None, 'domain': MultiAttr(), 'state': 'online', 'puppet_environment': None, 'graphite_graphs': MultiAttr(), 'function': None, 'security_availability': None, 'hypervisor': None, 'puppet_ca': 'ca.puppet.ig.local', 'hostname': None, 'allow_from': MultiAttr(), 'barman_host': None, 'network_zone': None, 'served_game': None, 'game_world': None, 'route_network': None, 'security_integrity': None, 'vlan_interfaces': MultiAttr(), 'ssh_users': MultiAttr(), 'primary_ip6': None, 'backup_freq': MultiAttr(), 'sshfp': MultiAttr(), 'backup_maxage': 36, 'game_market': None, 'subproject': None, 'fixup_enabled': False, 'igvm_locked': False, 'bladecenter': None})
`

igvm/host.py
HerrSpace: This gives you utc:
```python3
from datetime import datetime
datetime.utcnow().strftime("%s")
```

igvm/host.py
HerrSpace: I think so. I'd use None too.

HerrSpace: Probably best to also test this against testing serveradmin with a datetime attribute though.

igvm/host.py
HerrSpace: Ah also adminapi usually want native datatypes, not sure if we ever implemented it for datetime stuff though. We should create an attribute on serveradmin testing and try.

igvm/host.py
```diff
@@ -138,13 +139,13 @@ def put(self, remote_path, local_path, mode='0644'):
             )
 
     def acquire_lock(self, allow_fail=False):
-        if self.dataset_obj['igvm_locked']:
+        if bool(self.dataset_obj['igvm_locked']):
```
HerrSpace: This would be cast to bool implicitly anyhow. If you want to make it more explicit better check for `not None`

igvm/host.py
HerrSpace: I tested it, this is fine.

igvm/host.py
HerrSpace: So this is a problem. We do have a date datatype and it does take and give proper datetime objects on adminapi side, but we don't have a datetime datatype. We can't save time in it.

HerrSpace:
This change was picked up in https://github.com/innogames/igvm/pull/105

PR #102, opened 2019-03-11
WIP: Provide communication channel between VM and HV

TuxPowered42:


mtkh:
I think we have to first install `qemu-guest-agent` on all our domains for this to work.

TuxPowered42:
> I think we have to first install `qemu-guest-agent` on all our domains for this to work.

It's the opposite - quemu-guest-agent requires this to be on domain first.

TuxPowered42:
This MR won't be needed with the new IGVM Daemon implemmentation.

PR #106, opened 2019-04-02
Use sudo instead of root user when cleaning certs

Woellchen:


TuxPowered42:
https://github.com/innogames/igvm/pull/81 should provide the same but with admins' own accounts.

PR #108, opened 2019-04-05
Fix volume formatting race condition

HerrSpace:


igvm/hypervisor.py
```diff
@@ -399,6 +399,12 @@ def create_vm_storage(self, vm, transaction=None, vol_name=None):
         if transaction:
             transaction.on_rollback('destroy storage', volume.delete)
 
+        # XXX: When building a VM we use the volumes path to format it right
+        # after creation.  Unfortunately the kernel is slow to pick up on zfs
+        # volume changes and creates the symlink in /dev/zvol/<pool>/ only
+        # after a moment.
+        self.run("while [ ! -L '{}' ]; do sleep 1; done".format(volume.path()))
```
TuxPowered42: Can't we do it via libvirt calls somehow?

HerrSpace: Hey I almost missed this as this was already merged. I just had the tab still open.
I don't think so. My guess is based on volume.delete also failing during the rollback because it also tries to access the volume by path. I agree though that this fix (in.. not shitty) should probably be upstreamed and handled inside libvirt. Either .createXML() or .path() should block / timeout.

PR #109, opened 2019-05-09
Copy-edit the comments and the log on hypervisor_preferences.py

hasegeli:
On the previous log message, the number was one off, because the loop
is broken after hitting the not-executed preference.  The information
was also wrong, because it doesn't really mean that many preferences
are used to select this particular hypervisor.  It may be that sorting
this hypervisor before the next one caused the preferences to be
executed.  We can do better in here by comparing the number of executed
preferences for the next hypervisor on the line, or rolling out our own
sorting algorithm to be able to track this information, but I don't
think the information is worth the effort.  While cleaning this up,
I've also inlined the loop.

### DIFF
--- a/igvm/hypervisor_preferences.py
+++ b/igvm/hypervisor_preferences.py
@@ -1,4 +1,4 @@
-"""igvm - Hypervisor Preferences
+"""igvm - Hypervisor Selection Preferences
 
 This module contains preferences to select hypervisors.  Preferences return
 a value of any comparable datatype.  Only the return values of the same
@@ -6,13 +6,12 @@ preference is compared with each other.  Smaller values mark hypervisors
 as more preferred.  Keep in mind that for booleans false is less than true.
 See sorted_hypervisors() function below for the details of sorting.
 
-NOTE: This module contains the preferences as simple class form.  We try
-to keep them reusable, even though most of them are not reused.  Some of
-the classes are so simple, they could as well just be a function but kept
-as classes to have a consistent style.
-
 Copyright (c) 2018 InnoGames GmbH
 """
+# This module contains the preferences as simple classes.  We try to keep
+# them reusable, even though most of them are not reused.  Some of the classes
+# are so simple that they could as well just be a function, but kept
+# as classes to have a consistent style.
 
 from logging import getLogger
 
@@ -167,8 +166,8 @@ def sorted_hypervisors(preferences, vm, hypervisors):
     The most preferred ones will be yielded first.  The caller may then verify
     and use the hypervisors.  For sorting, we simply put the results
     of the preferences to any array for every hypervisor, and let Python
-    sort the arrays.  Unlikely semantically-low-level programming languages
-    like C, Python can compare arrays alright.  It would recursively compare
+    sort the arrays.  Unlike semantically-low-level programming languages
+    like C, Python can compare arrays just fine.  It would recursively compare
     the elements of the arrays with each other, and stop when they differ.
 
     As we know that most of the time it wouldn't be necessary to compare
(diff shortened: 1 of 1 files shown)


hasegeli:
I was reading this and noticed a minor issue.  It is not tested but unlikely to fail.

HerrSpace:
retest this please

HerrSpace:
ok to test

igvm/hypervisor_preferences.py
```diff
@@ -179,25 +178,23 @@ class to let them be prepared lazily.  When Python needs to compare
     executed for that hypervisor by LazyCompare.
 
     The LazyCompare optimization has one other benefit of bring more
```
HerrSpace: s/bring/bringing/

igvm/hypervisor_preferences.py
```diff
@@ -179,25 +178,23 @@ class to let them be prepared lazily.  When Python needs to compare
...
-        log.info(
-            'Hypervisor "{}" selected with decisive preference {!r} '
-            'after checking {} preferences.'
-            .format(hypervisor, comparable.func, index)
-        )
+        log.info('Hypervisor "{}" selected using {} preferences.'.format(
+            hypervisor, next(
+                (i for i, c in enumerate(comparables) if not c.executed), plen
+            )
+        ))
```
HerrSpace: Hm, good catch, though I think this makes it much harder to read. I'd prefer we keep the old loop style.
We can even show the correct preference via comparables[index-1], right?

igvm/hypervisor_preferences.py
```diff
@@ -179,25 +178,23 @@ class to let them be prepared lazily.  When Python needs to compare
...
-    visibility about selection.  After the sorting is done, we can check
+    visibility about the selection.  After the sorting is done, we can check
     which preferences are actually executed for the selected hypervisor,
-    and log which preference caused this hypervisor to be sorted before
-    the next one.
+    and log which preference caused this hypervisor to be sorted after
+    the previous or before the next one.
     """
-    log.debug('Sorting hypervisors by preference...')
+    plen = len(preferences)
```
HerrSpace: len() on lists is O(1). Let's just re-query it instead of having this obscurely named variable.

igvm/hypervisor_preferences.py
hasegeli: Reverted the loop.

We can still show the preference, but it doesn't mean much, because it is not the one that has necessarily caused this hypervisor to be sorted before the next one in line.

igvm/hypervisor_preferences.py
```diff
@@ -194,10 +194,11 @@ class to let them be prepared lazily.  When Python needs to compare
         for index, comparable in enumerate(comparables):
             if not comparable.executed:
                 break
+        else:
+            index += 1
```
HerrSpace: Just to make me remember, I see three scenarios here:
- The loop ran e.g. 3 times with indexes from 0-2 and we hit the break on index==2. We won't hit the else as we used break. index is now 2, which is correct. ✅
- The loop ran e.g. 3 times with indexes from 0-2 and we never hit the break. We will hit the else. Index is now 3, which is correct. ✅
- The loop runs 0 times if enumerate(comparables) is empty to begin with. We will hit the else. Index would be undefined here and this would throw an NameError. ❌

We would have hit the same NameError before once we hit the logger call below, so I won't make you fix this if you don't care. Can't happen for us in production currently anyway.

HerrSpace: Do you think it's worth changing this to?:
```python
try:
    index += 1
except NameError:
    index = 0
```

HerrSpace:
lgtm, thanks!

igvm/hypervisor_preferences.py
hasegeli: `comparables` cannot really be empty, but I've fixed it anyway.  `len()` is maybe also slightly more self-explanatory.

HerrSpace: Ah, that's a great idea. +1

PR #110, opened 2019-05-15
Remove rootuser for puppet access

eviscares:
Root access should be disabled everywhere, so we shouldn't rely on it.


HerrSpace:
Duplicate of https://github.com/innogames/igvm/pull/106 and https://github.com/innogames/igvm/pull/81

PR #113, opened 2019-06-26
Use xfs reflog on Debian Stretch

TuxPowered42:


igvm/settings.py
```diff
@@ -63,6 +63,10 @@
     ],
 }
 
+XFS_CONFIG = {
+    'jessie': [],
+    'stretch': ['-m reflink=1'],
```
seqizz: Can buster also haz some love?

TuxPowered42: Sure. To be honest I'd prefer to add it only on Buster to catch bugs if there are any. I was not really able to tell if Stretch kernel support for those things is "stable".

PR #114, opened 2019-07-05
Remove unused __repr__()s from hypervisor_preferences.py

hasegeli:


hasegeli:
I was reading this again and noticed those now being unneeded.  It is not tested but unlikely to fail.

HerrSpace:
Why remove the hash difference? It seems like a sane idea.

HerrSpace:
retest this please

hasegeli:
`sorted_hypervisors()` sorts `([preference, ...], hypervisor)`s tuples.  I think the hypervisors would be sorted stably if all preferences happen to return the same value.

HerrSpace:
The tests fail by raising a bunch of these:
```
        # Use decorate-sort-undecorate pattern to log details about sorting
        for comparables, hypervisor in sorted(
            ([LazyCompare(p, vm, h) for p in preferences], h)
>           for h in hypervisors
        ):
E       TypeError: unorderable types: Hypervisor() < Hypervisor()

igvm/hypervisor_preferences.py:151: TypeError
```

hasegeli:
Oh, I missed that the objects don't compare by default on Python 3.  I removed that commit from the request.

HerrSpace:
This seems somewhat pointless to me now. Having the __repr__s is kinda usefull for printf debugging during development.

PR #117, opened 2019-08-12
TECH-24783: use libvirt data, not system-wide in HYPERVISOR_PREFERENCES

TuxPowered42:
Use memory and storage pool allocation information taken from libvirt
(and cached by Graphite and Serveradmiin) into account when choosing
hypervisor for build and migration target. Replace memory and
disk_size_gib attributes with libvirt_(pool|memory)_total_gib when
calculating insufficient resources and use
libvirt_(pool|memory)_free_gib when sorting hypervisors.

This should make searching for usable HV faster.

### DIFF
--- a/igvm/hypervisor_preferences.py
+++ b/igvm/hypervisor_preferences.py
@@ -22,23 +22,32 @@ log = getLogger(__name__)
 
 class InsufficientResource(object):
     """Check a resource of hypervisor would be sufficient"""
-    def __init__(self, attribute, reserved=0):
-        self.attribute = attribute
+    def __init__(self, hv_attribute, vm_attribute, multiplier=1, reserved=0):
+        self.hv_attribute = hv_attribute
+        self.vm_attribute = vm_attribute
+        self.multiplier = multiplier # TODO: Use identical units in Serveradmin.
         self.reserved = reserved
 
     def __repr__(self):
-        args = repr(self.attribute)
+        args = repr(self.hv_attribute)
         if self.reserved:
             args += ', reserved=' + repr(self.reserved)
 
         return '{}({})'.format(type(self).__name__, args)
 
     def __call__(self, vm, hv):
-        total_size = hv.dataset_obj[self.attribute]
-        vms_size = sum(v[self.attribute] for v in hv.dataset_obj['vms'])
+        # Treat freshly created HVs always passing this check
+        if not hv.dataset_obj[self.hv_attribute]:
+            return False
+
+        total_size = hv.dataset_obj[self.hv_attribute]
+        vms_size = sum(
+            vm[self.vm_attribute] * self.multiplier
+            for vm in hv.dataset_obj['vms']
+        )
         remaining_size = total_size - vms_size - self.reserved
 
-        return remaining_size < vm.dataset_obj[self.attribute]
+        return remaining_size < vm.dataset_obj[self.vm_attribute]
 
 
 class OtherVMs(object):
--- a/igvm/settings.py
+++ b/igvm/settings.py
@@ -104,13 +104,15 @@ IMAGE_PATH = '/tmp'
 HYPERVISOR_ATTRIBUTES = [
     'cpu_util_pct',
     'cpu_util_vm_pct',
-    'disk_size_gib',
     'hardware_model',
     'hostname',
     'igvm_locked',
     'intern_ip',
     'iops_avg',
-    'memory',
+    'libvirt_memory_total_gib',
+    'libvirt_memory_used_gib',
+    'libvirt_pool_total_gib',
+    'libvirt_pool_used_gib',
     'num_cpu',
     'os',
     'state',
@@ -166,9 +168,8 @@ VM_ATTRIBUTES = [
 # preference is only going to be checked when the previous ones return all
 # the same values.
 HYPERVISOR_PREFERENCES = [
-    # We assume 10 GiB for root partition, 16 for swap, and 6 reserved.
-    InsufficientResource('disk_size_gib', reserved=32),
-    InsufficientResource('memory'),
+    InsufficientResource('libvirt_pool_total_gib', 'disk_size_gib'),
+    InsufficientResource('libvirt_memory_total_gib', 'memory', multiplier=1024),
     # Checks the maximum vCPU usage (95 percentile) of the given hypervisor
     # for the given time_range and dismisses it as target when it is over
     # the value of threshold.


TuxPowered42:
retest this please

TuxPowered42:
retest this, please

TuxPowered42:
retest this, please

igvm/hypervisor_preferences.py
```diff
@@ -22,23 +22,28 @@
...
+        args = repr(self.hv_attribute)
         if self.reserved:
             args += ', reserved=' + repr(self.reserved)
 
         return '{}({})'.format(type(self).__name__, args)
 
     def __call__(self, vm, hv):
-        total_size = hv.dataset_obj[self.attribute]
-        vms_size = sum(v[self.attribute] for v in hv.dataset_obj['vms'])
+        # Treat freshly created HVs always passing this check
```
HerrSpace: isn't returning False equivalent to failing this check?

HerrSpace: Check is probably the right thing to do anyway.

igvm/settings.py
```diff
@@ -193,6 +194,10 @@
     HypervisorAttributeValue('cpu_util_pct'),
     # Find Hypervisor with less I/O utilization
     HypervisorAttributeValue('iops_avg'),
+    # Find hypervisor with less used memory
+    HypervisorAttributeValue('libvirt_pool_used_gib'),
+    # Find hypervisor with less used disk
+    HypervisorAttributeValue('libvirt_memory_used_gib'),
```
HerrSpace: I wouldn't add these.
- They don't make sense if your pools aren't of the same size.
- The OtherVMs below probably still have more value then these.
- The HYPERVISOR_PREFERENCES mechanism is much to dumb to properly optimize placement anyway. You'd have to shift VMs around based on multiple attributes.

igvm/settings.py
```diff
@@ -166,9 +168,8 @@
 # preference is only going to be checked when the previous ones return all
 # the same values.
 HYPERVISOR_PREFERENCES = [
-    # We assume 10 GiB for root partition, 16 for swap, and 6 reserved.
-    InsufficientResource('disk_size_gib', reserved=32),
-    InsufficientResource('memory'),
+    InsufficientResource('libvirt_pool_total_gib', 'disk_size_gib'),
+    InsufficientResource('libvirt_memory_total_gib', 'memory'),
```
HerrSpace: this is sweet!

TuxPowered42: Anything to resolve here?

igvm/settings.py
TuxPowered42: That was ticket from you to sort HVs by free storage pool and memory. Unfortunately this function HypervisorAttributeValue sorts in way which makes sense for "used" and not "free" resource.

igvm/settings.py
HerrSpace: Nah

igvm/settings.py
HerrSpace: It just doesn't make sense like this though and the InsufficientResource checks on top should be enough anyhow, if the fields cached from libvirt aren't stale. If they are stale, this won't help either.

igvm/hypervisor_preferences.py
```diff
@@ -22,23 +22,28 @@
...
+        if not hv.dataset_obj[self.hv_attribute]:
+            return False
```
HerrSpace: Why pass and not fail here? You'll always have enough proper once with graphite data anyway and if you don't, this isn't a nock-out criteria anyway. Failing this by returning True seems to me the more safe thing to do, no?

TuxPowered42: A freshly installed HV should be usable immediately, not after a few hours only when cron job on Serveradmin runs.

igvm/settings.py
TuxPowered42: InsufficientResource also uses cached values and might still return HVs which are in fact not usable, even if we use values acquired in better way (from libvirt). It was never about optimal placement, it was about finding an usab le HV faster, as last check is done by locking a HV and getting real data from it.

igvm/hypervisor_preferences.py
HerrSpace: Hmm. Fair enough.

igvm/settings.py
HerrSpace: Let me frame it differently.
In the cases where we actually run this check we _might_ shave off between one and five hypervisor disk checks. Connecting to each and checking for the disks takes maybe .5 second. So best case we just saved 2.5 seconds. In a 5 to 15 minute igvm build/migrate. In the average run we probably save around 0 seconds.
On the downside we introduce code which makes little sense if your lvm pools aren't all of the same size.

igvm/hypervisor_preferences.py
kofrezo: It would be usable but less likey be choosen as you return False here no ?

igvm/settings.py
kofrezo: We have used graphite cached data attributes from serveradmin in igbalance before. The problem was exactly what you described above 

> TuxPowered42 15 days ago Author Member
> A freshly installed HV should be usable immediately, not after a few hours only when cron job on Serveradmin runs.

The information is outdated rather fast. So imagine you are evacuating a hypervisor or setup a fresh one. It becomes a very high rated candidate and will likely be on the top of the list. Problem is that after 1 or 2 migrations (or builds) the graphite information is outdated unless the graphite cache in serveradmin is updated. This may take some time as you already pointed out above. This hypervisor will be choosen over and over again and only get dismissed at the preflight check. The pre flight check is a very expensive operation requiring us to SSH to the hypervisor which is why we move it to the very end and try to execute it only if really needed.

This said I think you will not achieve what you expect. In single cases where we migrate or build single VMs or have a lot of hardware it will bring speed but for example for upcoming buster migration in areas with less hardware (Think of VN) and many migrations and builds in a row it will actully slow us down.

igvm/hypervisor_preferences.py
TuxPowered42: This is not for sorting preference but for having HV in result or not having it at all. Function is called "InsufficientResource" and False means that the resource is not insufficient.

igvm/settings.py
TuxPowered42: @HerrSpace I lack real numbers but in my memory it is slower than half second per HV and there might be tens of HVs to check. For migrating of some small machines this adds up substantially.

TuxPowered42: @kofrezo The information is outdated, that is true. But this moment we must check even more HVs, if we could skip just a few, we would gain some time. If you still don't agree then let's remove this part, work on Serveradmin cache speed and then reintroduce it.

TuxPowered42: This part of change was removed from my PR.

PR #119, opened 2019-08-29
Add reserved memory to HV selection

Phantuman:
We want to take the reserved memory into consideration while selecting Hypervisors.
Else we will check a lot of hypervisors which will then reject building the vm on them


HerrSpace:
@Phantuman doesn't seem to be in the admins team, that's why the tests aren't run. Add him and somebody from the admin team type 'test this please' here to run the tests once.

HerrSpace:
Also this would be superseded by https://github.com/innogames/igvm/pull/117/files anyhow, no?

PR #120, opened 2019-09-16
Add AWS support for common igvm functions

lacunoc:
This adds the basic functionality to be able to build, delete (incl. retire in SA), start, stop and restart VMs in AWS.

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -66,7 +66,8 @@ def evacuate(hv_hostname, offline=None, dry_run=False):
     with _get_hypervisor(hv_hostname, allow_reserved=True) as hv:
         if dry_run:
             log.info('I would set {} to state online reserved'.format(
-                hv_hostname))
+                hv_hostname)
+            )
         else:
             hv.dataset_obj['state'] = 'online_reserved'
             hv.dataset_obj.commit()
@@ -99,8 +100,8 @@ def vcpu_set(vm_hostname, count, offline=False):
 
         if offline and not vm.is_running():
             log.info(
-                '"{}" is already powered off, ignoring --offline.'
-                .format(vm.fqdn)
+                '"{}" is already powered off, ignoring --offline.'.format(
+                    vm.fqdn)
             )
             offline = False
 
@@ -140,8 +141,8 @@ def mem_set(vm_hostname, size, offline=False):
 
         if offline and not vm.is_running():
             log.info(
-                '"{}" is already powered off, ignoring --offline.'
-                .format(vm.fqdn)
+                '"{}" is already powered off, ignoring --offline.'.format(
+                    vm.fqdn)
             )
             offline = False
 
@@ -319,6 +320,7 @@ def vm_migrate(vm_hostname, hypervisor_hostname=None,
 
             def _reset_hypervisor():
                 vm.hypervisor = previous_hypervisor
+
             transaction.on_rollback('reset hypervisor', _reset_hypervisor)
 
             if run_puppet:
@@ -422,15 +424,15 @@ def vm_delete(vm_hostname, retire=False):
             vm.dataset_obj['state'] = 'retired'
             vm.dataset_obj.commit()
             log.info(
-                '"{}" is destroyed and set to "retired" state.'
-                .format(vm.fqdn)
+                '"{}" is destroyed and set to "retired" state.'.format(
+                    vm.fqdn)
             )
         else:
             vm.dataset_obj.delete()
             vm.dataset_obj.commit()
             log.info(
-                '"{}" is destroyed and deleted from Serveradmin'
-                .format(vm.fqdn)
+                '"{}" is destroyed and deleted from Serveradmin'.format(
+                    vm.fqdn)
             )
 
 
@@ -456,8 +458,8 @@ def vm_sync(vm_hostname):
         if changed:
             vm.dataset_obj.commit()
             log.info(
-                '"{}" is synchronized {} attributes ({}).'
-                .format(vm.fqdn, len(changed), ', '.join(changed))
+                '"{}" is synchronized {} attributes ({}).'.format(
+                    vm.fqdn, len(changed), ', '.join(changed))
             )
         else:
             log.info(
@@ -465,7 +467,7 @@ def vm_sync(vm_hostname):
             )
 
 
-@with_fabric_settings   # NOQA: C901
+@with_fabric_settings  # NOQA: C901
 def host_info(vm_hostname):
     """Extract runtime information about a VM
 
@@ -513,15 +515,14 @@ def host_info(vm_hostname):
             simple_stats = (
                 'Current: {} {unit}\n'
                 'Free:    {} {unit}\n'
-                'Max:     {} {unit}'
-                .format(capacity - free, free, capacity, unit=unit)
-            )
+                'Max:     {} {unit}'.format(
+                    capacity - free, free, capacity, unit=unit))
 
             if not 0 <= free <= capacity > 0:
                 log.warning(
                     '{} ({}) and {} ({}) have weird ratio, skipping progress '
-                    'calculation'
-                    .format(free_key, free, capacity_key, capacity)
+                    'calculation'.format(
+                        free_key, free, capacity_key, capacity)
                 )
                 info[result_key] = red(simple_stats)
                 return
@@ -538,8 +539,7 @@ def host_info(vm_hostname):
             max_bars = 20
             num_bars = int(round(ratio * max_bars))
             info[result_key] = (
-                '[{}{}] {}%\n{}'
-                .format(
+                '[{}{}] {}%\n{}'.format(
                     color('#' * num_bars), ' ' * (max_bars - num_bars),
                     int(round(ratio * 100)),
                     simple_stats,
@@ -696,15 +696,15 @@ def _get_best_hypervisor(vm, hypervisor_states, offline=False):
         except libvirtError as error:
             hypervisor.release_lock()
             log.warning(
-                'Preferred hypervisor "{}" is skipped: {}'
-                .format(hypervisor, error)
+                'Preferred hypervisor "{}" is skipped: {}'.format(
+                    hypervisor, error)
             )
             continue
         except HypervisorError as error:
             hypervisor.release_lock()
             log.warning(
-                'Preferred hypervisor "{}" is skipped: {}'
-                .format(hypervisor, error)
+                'Preferred hypervisor "{}" is skipped: {}'.format(
+                    hypervisor, error)
             )
             continue
 
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -12,6 +12,7 @@ from adminapi.filters import Any, StartsWith
 from fabric.colors import green, red, white, yellow
 from fabric.network import disconnect_all
 from ipaddress import ip_address
+from jinja2 import Environment, FileSystemLoader
 from libvirt import libvirtError
 
 from igvm.exceptions import (
@@ -25,6 +26,8 @@ from igvm.host import with_fabric_settings
 from igvm.hypervisor import Hypervisor
 from igvm.hypervisor_preferences import sorted_hypervisors
 from igvm.settings import (
+    AWS_CONFIG,
+    AWS_RETURN_CODES,
     HYPERVISOR_ATTRIBUTES,
     HYPERVISOR_PREFERENCES,
     VM_ATTRIBUTES,
@@ -96,6 +99,12 @@ def vcpu_set(vm_hostname, count, offline=False):
     with ExitStack() as es:
         vm = es.enter_context(_get_vm(vm_hostname))
 
+        if vm.dataset_obj['igvm_operation_mode'] != 'kvm':
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
+
         _check_defined(vm)
 
         if offline and not vm.is_running():
@@ -126,6 +135,12 @@ def mem_set(vm_hostname, size, offline=False):
     """
     with ExitStack() as es:
         vm = es.enter_context(_get_vm(vm_hostname))
+
+        if vm.dataset_obj['igvm_operation_mode'] != 'kvm':
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
 
         _check_defined(vm)
 
@@ -166,6 +181,12 @@ def disk_set(vm_hostname, size):
     with ExitStack() as es:
         vm = es.enter_context(_get_vm(vm_hostname))
 
+        if vm.dataset_obj['igvm_operation_mode'] != 'kvm':
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
+
         _check_defined(vm)
 
         current_size_gib = vm.dataset_obj['disk_size_gib']
@@ -197,6 +218,12 @@ def change_address(vm_hostname, new_address, offline=False):
         raise IGVMError('IP address change can be only performed offline')
 
     with _get_vm(vm_hostname) as vm:
+        if vm.dataset_obj['igvm_operation_mode'] != 'kvm':
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
+
         old_address = vm.dataset_obj['intern_ip']
         new_address = ip_address(new_address)
 
@@ -236,45 +263,79 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
     with ExitStack() as es:
         vm = es.enter_context(_get_vm(vm_hostname))
 
-        if vm.hypervisor:
-            es.enter_context(_lock_hv(vm.hypervisor))
-        else:
-            vm.hypervisor = es.enter_context(_get_best_hypervisor(
-                vm,
-                ['online', 'online_reserved'] if allow_reserved_hv
-                else ['online'],
-                True,
-            ))
-            vm.dataset_obj['hypervisor'] = \
-                vm.hypervisor.dataset_obj['hostname']
+        if vm.dataset_obj['igvm_operation_mode'] == 'aws':
+            file_loader = FileSystemLoader('igvm/templates/')
+            env = Environment(loader=file_loader)
+            template = env.get_template('aws_user_data.cfg')
 
-        if vm.hypervisor.vm_defined(vm) and vm.is_running():
-            raise InvalidStateError(
-                '"{}" is still running.'.format(vm.fqdn)
+            user_data = template.render(
+                hostname=vm.dataset_obj['hostname'].rstrip('.ig.local'),
+                fqdn=vm.dataset_obj['hostname'],
+                apt_repos=AWS_CONFIG[0]['apt'],
+                puppet_master_addr=AWS_CONFIG[0]['puppet']['master_addr'],
+                puppet_master=vm.dataset_obj['puppet_master'],
+                puppet_ca=vm.dataset_obj['puppet_ca'],
+                puppet_ca_addr=AWS_CONFIG[0]['puppet']['ca_addr'],
             )
 
-        if rebuild and vm.hypervisor.vm_defined(vm):
-            vm.hypervisor.undefine_vm(vm)
+            vm.aws_build(
+                run_puppet=run_puppet,
+                debug_puppet=debug_puppet,
+                postboot=user_data
+            )
+        elif vm.dataset_obj['igvm_operation_mode'] == 'kvm':
+            if vm.hypervisor:
+                es.enter_context(_lock_hv(vm.hypervisor))
+            else:
+                vm.hypervisor = es.enter_context(_get_best_hypervisor(
+                    vm,
+                    ['online', 'online_reserved'] if allow_reserved_hv
+                    else ['online'],
+                    True,
+                ))
+                vm.dataset_obj['hypervisor'] = \
+                    vm.hypervisor.dataset_obj['hostname']
 
-        vm.build(
-            run_puppet=run_puppet,
-            debug_puppet=debug_puppet,
-            postboot=postboot,
-        )
+            if vm.hypervisor.vm_defined(vm) and vm.is_running():
+                raise InvalidStateError(
+                    '"{}" is still running.'.format(vm.fqdn)
+                )
+
+            if rebuild and vm.hypervisor.vm_defined(vm):
+                vm.hypervisor.undefine_vm(vm)
+
+            vm.build(
+                run_puppet=run_puppet,
+                debug_puppet=debug_puppet,
+                postboot=postboot,
+            )
+        else:
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
 
         vm.dataset_obj.commit()
 
 
-@with_fabric_settings   # NOQA: C901
+@with_fabric_settings  # NOQA: C901
 def vm_migrate(vm_hostname, hypervisor_hostname=None,
                run_puppet=False, debug_puppet=False,
                offline=False, offline_transport='drbd',
                allow_reserved_hv=False, no_shutdown=False):
     """Migrate a VM to a new hypervisor."""
+
     with ExitStack() as es:
         vm = es.enter_context(
             _get_vm(vm_hostname, allow_retired=True)
         )
+
+        if vm.dataset_obj['igvm_operation_mode'] != 'kvm':
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
+
         if hypervisor_hostname:
             hypervisor = es.enter_context(_get_hypervisor(
                 hypervisor_hostname, allow_reserved=allow_reserved_hv
@@ -345,28 +406,43 @@ def vm_migrate(vm_hostname, hypervisor_hostname=None,
 def vm_start(vm_hostname):
     """Start a VM"""
     with _get_vm(vm_hostname) as vm:
-        _check_defined(vm)
-
-        if vm.is_running():
-            log.info('"{}" is already running.'.format(vm.fqdn))
-            return
-        vm.start()
+        if vm.dataset_obj['igvm_operation_mode'] == 'aws':
+            vm.aws_start()
+        elif vm.dataset_obj['igvm_operation_mode'] == 'kvm':
+            _check_defined(vm)
+            if vm.is_running():
+                log.info('"{}" is already running.'.format(vm.fqdn))
+                return
+            vm.start()
+        else:
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
 
 
 @with_fabric_settings
 def vm_stop(vm_hostname, force=False):
     """Gracefully stop a VM"""
     with _get_vm(vm_hostname, allow_retired=True) as vm:
-        _check_defined(vm)
+        if vm.dataset_obj['igvm_operation_mode'] == 'aws':
+            vm.aws_shutdown()
+        elif vm.dataset_obj['igvm_operation_mode'] == 'kvm':
+            _check_defined(vm)
 
-        if not vm.is_running():
-            log.info('"{}" is already stopped.'.format(vm.fqdn))
-            return
-        if force:
-            vm.hypervisor.stop_vm_force(vm)
+            if not vm.is_running():
+                log.info('"{}" is already stopped.'.format(vm.fqdn))
+                return
+            if force:
+                vm.hypervisor.stop_vm_force(vm)
+            else:
+                vm.shutdown()
+            log.info('"{}" is stopped.'.format(vm.fqdn))
         else:
-            vm.shutdown()
-        log.info('"{}" is stopped.'.format(vm.fqdn))
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode'])
+            )
 
 
 @with_fabric_settings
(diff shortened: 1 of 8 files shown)


lacunoc:
retest this please

igvm/settings.py
```diff
@@ -162,6 +170,19 @@
...
+    'aws-fra',
+]
+
+AWS_RETURN_CODES = {
+    'pending': '0',
+    'running': '16',
+    'shutting-down': '32',
+    'terminated': '48',
+    'stopping': '64',
+    'stopped': '80',
```
TuxPowered42: Shouldn't those be integers? Those are bit flags after all.

igvm/settings.py
```diff
@@ -162,6 +170,19 @@
     {'hypervisor': HYPERVISOR_ATTRIBUTES},
 ]
 
+AWS_DATACENTERS = [
+    'aws-fra',
```
TuxPowered42: Wouldn't it be better to make this an attribute of a network in Serveradmin?

igvm/templates/user_data.cfg
```diff
@@ -0,0 +1,92 @@
+#cloud-config
```
TuxPowered42: user_data.cfg seems a bit generic to me.

igvm/templates/user_data.cfg
```diff
@@ -0,0 +1,92 @@
...
+package_upgrade: true
+packages:
+ - puppet-agent
+ - puppet-msgpack
+
+runcmd:
+  - echo "# We need to use external IPs for puppet here" >> /etc/hosts
+  - echo 212.53.146.132 {{ puppet_ca }} >> /etc/hosts
+  - echo 212.53.146.131 {{ puppet_master }} >> /etc/hosts
+  - /opt/puppetlabs/puppet/bin/puppet agent --detailed-exitcodes --fqdn={{ fqdn }} --server={{ puppet_master }} --ca_server={{ puppet_ca }} --no-report --waitforcert=60 --onetime --no-daemonize --verbose
```
TuxPowered42: Can you make it fit in 80 columns?

igvm/templates/user_data.cfg
```diff
@@ -0,0 +1,92 @@
...
+
+final_message: "The system is finally up, after $UPTIME seconds"
```
TuxPowered42: Newline at the end?

igvm/vm.py
```diff
@@ -197,6 +201,47 @@ def start(self, force_stop_failed=True, transaction=None):
...
+
+        ec2 = boto3.client('ec2')
+
+        try:
+            ec2.start_instances(
+                InstanceIds=[self.dataset_obj['aws_instance_id']],
+                DryRun=True
+            )
+        except ClientError as e:
+            if 'DryRunOperation' not in str(e):
```
TuxPowered42: Can this information really be only obtained by parsing a string?

igvm/vm.py
```diff
@@ -208,6 +253,96 @@ def shutdown(self, check_vm_up_on_transaction=True, transaction=None):
...
+        """
+
+        ec2 = boto3.client('ec2')
+
+        try:
+            response = ec2.terminate_instances(
+                InstanceIds=[self.dataset_obj['aws_instance_id']])
+            log.info(response)
+        except ClientError as e:
+            log.error(e)
```
TuxPowered42: No re-raise as igvm exception?

igvm/vm.py
```diff
@@ -335,6 +469,95 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
...
+                            },
+                        ]
+                    },
+                ],
+                DryRun=False,
+                MinCount=1,
+                MaxCount=1)
+            log.info(response)
+        except ClientError as e:
+            raise ClientError(e)
```
TuxPowered42: Shouldn't this be casted to igvm exception?

igvm/vm.py
```diff
@@ -335,6 +469,95 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
...
+
+        if run_puppet:
```
TuxPowered42: How do you run Puppet before waiting for VM start, cloud init and whatnot? This is AWS VM without chroot access, so puppet is run on already running VM.

igvm/vm.py
```diff
@@ -197,6 +201,47 @@ def start(self, force_stop_failed=True, transaction=None):
         if transaction:
             transaction.on_rollback('stop VM', self.shutdown)
 
+    def aws_start(self):
```
TuxPowered42: This function is never called. So you provided only low level api. Also is it aws_start or aws_build?

igvm/settings.py
lacunoc: Unfortunately the API is assuming these to be strings
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter Filters[0].Values[0], value: 0, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[1], value: 16, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[2], value: 32, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[3], value: 48, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[4], value: 64, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[5], value: 80, type: <class 'int'>, valid types: <class 'str'>
```

igvm/templates/user_data.cfg
lacunoc: In AWS console its called `user_data` and the API parameter is `UserData`. Should I prefix it with `aws_` ?

lacunoc: Or maybe `aws_cloudinit.cfg` ?

igvm/settings.py
lacunoc: I made them Integers and only cast them to String when calling `aws_describe_instance_status`. This makes also return code comparison a lot easier

igvm/templates/user_data.cfg
lacunoc: Looks ugly, bit now it's fitting 80 columns.

igvm/templates/user_data.cfg
lacunoc: Newline added

igvm/vm.py
lacunoc: VMError is now raised

igvm/vm.py
lacunoc: VMError is now raised

igvm/vm.py
lacunoc: A `puppet_run` for a VM located in `AWS_DATACENTERS` is only removing the cert on puppet ca server during build.

igvm/vm.py
lacunoc: Its called in `vm_build` in commands.py and it is starting an already build VM in AWS.

igvm/vm.py
lacunoc: Yes, it can:

> Parameters
> DryRun (boolean) -- Checks whether you have the required permissions for the action, without actually making the request, and provides an error response. If you have the required permissions, the error response is DryRunOperation . Otherwise, it is UnauthorizedOperation .

Reference: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html

igvm/vm.py
```diff
@@ -208,6 +252,100 @@ def shutdown(self, check_vm_up_on_transaction=True, transaction=None):
...
+        :return: return code of instance state as int
+        """
+
+        ec2 = boto3.client('ec2')
+        response = ec2.describe_instances(
+            Filters=[
+                {
+                    'Name': 'instance-state-code',
+                    'Values': [
+                        str(AWS_RETURN_CODES['pending']),
```
kofrezo: I know at the end it gets consumed by the AWS web API but isn't it still better to use int here instead of str as we are dealing with numbers.

kofrezo:
Please mind that there would be also great value in having integration tests for the AWS VMs.

igvm/templates/user_data.cfg
```diff
@@ -0,0 +1,95 @@
...
+
+package_update: true
+package_upgrade: true
+packages:
+ - puppet-agent
+ - puppet-msgpack
+
+runcmd:
+  - echo "# We need to use external IPs for puppet here" >> /etc/hosts
+  - echo 212.53.146.132 {{ puppet_ca }} >> /etc/hosts
```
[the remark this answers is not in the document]
lacunoc: Because the name and address is not fitting. We have to use the internal name but the external IP address to be able to use puppet here.

igvm/vm.py
lacunoc: Afair the API call will fail, because, for whatever reason, it's expecting strings as instance-state-code. But the documentation is also telling me integer should be possible. I will double check

igvm/vm.py
lacunoc: We are hiding the DryRunOperation, but will raise UnauthorizedOperation e.g.

igvm/templates/user_data.cfg
lacunoc: renamed it to `aws_user_data.cfg`

igvm/settings.py
lacunoc: Shouldn't it be handled like every other datacenter? Then it should stick to the VM

igvm/vm.py
lacunoc: Yep, not working, thats why we need to use `str` here:
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter Filters[0].Values[0], value: 0, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[1], value: 16, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[2], value: 32, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[3], value: 48, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[4], value: 64, type: <class 'int'>, valid types: <class 'str'>
Invalid type for parameter Filters[0].Values[5], value: 80, type: <class 'int'>, valid types: <class 'str'>
```

igvm/templates/user_data.cfg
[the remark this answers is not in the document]
TuxPowered42: Unless @Woellchen can fix it :)

igvm/settings.py
TuxPowered42: Attribute of VM can be inherited (Serveradmin attribute inheritance) from network, that is not a problem. I only say that this should be a property in Serveradmin, not a hardcoded thing in IGVM.

igvm/templates/user_data.cfg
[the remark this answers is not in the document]
Woellchen: not in near future :) there are tickets for it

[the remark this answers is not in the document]
lacunoc: So this is out of scope of this PR and we will stick with this for now.

igvm/settings.py
lacunoc: I removed `AWS_DATACENTERS` settings and introduced `igvm_operation_mode` attribute in SA with this regex `\A(aws|azure|esxi|hyperv|kvm)\Z`
I also put `kvm` (ig-chroot) as an elif in the common functions, for adding more in the future easily. Furthermore a `NotImplementedError` will be raised for every not supported virtualization/provisioning method.

igvm/commands.py
```diff
@@ -235,45 +255,74 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
...
+        if vm.dataset_obj['igvm_operation_mode'] == 'aws':
+            file_loader = FileSystemLoader('igvm/templates/')
+            env = Environment(loader=file_loader)
+            template = env.get_template('aws_user_data.cfg')
 
-        if vm.hypervisor.vm_defined(vm) and vm.is_running():
-            raise InvalidStateError(
-                '"{}" is still running.'.format(vm.fqdn)
+            user_data = template.render(
+                hostname=vm.dataset_obj['hostname'].rstrip('.ig.local'),
```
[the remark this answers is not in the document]
lacunoc: Because beside of the FQDN we also need the hostname here within the template.

igvm/vm.py
```diff
@@ -335,6 +472,95 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
...
+        ec2 = boto3.client('ec2')
+
+        try:
+            response = ec2.run_instances(
+                ImageId=self.dataset_obj['aws_image_id'],
+                InstanceType=self.dataset_obj['aws_instance_type'],
+                KeyName=self.dataset_obj['aws_key_name'],
+                SecurityGroupIds=list(
+                    self.dataset_obj['aws_security_group_ids']),
+                SubnetId=self.dataset_obj['aws_subnet_id'],
```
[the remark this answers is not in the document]
lacunoc: True! It is planned to move it to the network objects when they are available.

igvm/vm.py
```diff
@@ -423,38 +647,38 @@ def run_puppet(self, clear_cert=False, debug=False):
...
+                '--detailed-exitcodes '
+                '--fqdn={} --server={} --ca_server={} '
+                '--no-report --waitforcert=60 --onetime --no-daemonize '
+                '--skip_tags=chroot_unsafe --verbose{} ) ;'
+                '[ $? -eq 2 ]'.format(
+                    self.fqdn,
+                    self.dataset_obj['puppet_master'],
+                    self.dataset_obj['puppet_ca'],
+                    ' --debug' if debug else '',
+                )
```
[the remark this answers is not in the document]
lacunoc: \o/

igvm/commands.py
```diff
@@ -75,8 +77,8 @@ def evacuate(hv_hostname, offline=None, dry_run=False):
             vm_function = vm['function']
 
             if (
-                offline is not None and
-                (offline == [] or vm_function in offline)
+                    offline is not None and
+                    (offline == [] or vm_function in offline)
```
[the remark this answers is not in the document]
lacunoc: reverted

igvm/commands.py
```diff
@@ -235,45 +255,74 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
...
+
+            vm.build(
+                run_puppet=run_puppet,
+                debug_puppet=debug_puppet,
+                postboot=postboot,
+            )
+        else:
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    vm.dataset_obj['igvm_operation_mode']))
```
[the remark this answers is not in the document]
lacunoc: I willl leave it like it is for now.

igvm/commands.py
```diff
@@ -680,7 +775,7 @@ def _get_best_hypervisor(vm, hypervisor_states, offline=False):
     }, HYPERVISOR_ATTRIBUTES))
 
     for hypervisor in sorted_hypervisors(
-        HYPERVISOR_PREFERENCES, vm, hypervisors
+            HYPERVISOR_PREFERENCES, vm, hypervisors
```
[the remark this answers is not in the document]
lacunoc: reverted

igvm/kvm.py
```diff
@@ -220,8 +212,8 @@ def _live_repin_cpus(domain, props, max_phys_cpus):
 
 
 def migrate_background(
-    domain, source, destination,
-    migrate_params, migrate_flags,
+        domain, source, destination,
+        migrate_params, migrate_flags,
```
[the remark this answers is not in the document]
lacunoc: reverted

igvm/templates/aws_user_data.cfg
```diff
@@ -0,0 +1,96 @@
...
+package_update: true
+package_upgrade: true
+packages:
+ - puppet-agent
+ - puppet-msgpack
+
+runcmd:
+  - echo "# We need to use external IPs for puppet here" >> /etc/hosts
+  - echo 212.53.146.132 {{ puppet_ca }} >> /etc/hosts
+  - echo 212.53.146.131 {{ puppet_master }} >> /etc/hosts
```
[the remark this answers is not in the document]
lacunoc: It's now coming from settings.py

igvm/templates/aws_user_data.cfg
```diff
@@ -0,0 +1,96 @@
...
+        iAsfEK51ED4QOfnRRYOlbBvDwL3XQBZtFVtznwI1DLg/JNVVXFM8sDei+eL3O5yo
+        OVEuTBRvNH7ONlFRcBdFCdfSi1VvKVY2JEQX4R4BHyZvD/B2pcBeTM5puav5QuwK
+        XppwHg4dBbs1BrRsHWGMdqTJWGkwwVUdt3af6WSWD6BA+rv2I3DM7BMt6t/FK1BM
+        apHGY2fj3gK8pWyi13R8tryr4ZdIa/RnwxWt8ySWclgDVSPxTcNE1Zn2fYDB/Ykv
+        TyF52dn3+6UW6mnjL45Hictqo9A6aOxre5dlwEyHXhurs/7UP0Ell20UEN1VsKDP
+        E2V8pq+JFXQ+F2sLMEar4L5teopk78L7t0FibqmKcFRVgvseBbjm0xHfdTk4mXCD
+        wbaBtX+E9qhd5e4brEYNx9I9I2fvXGw5sX6Qdty9HnEQOKGMshEAfl2iCzJ56n4C
+        AkPnU50v
+        =KPbA
+        -----END PGP PUBLIC KEY BLOCK-----
```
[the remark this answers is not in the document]
lacunoc: It's now also coming from settings.py

igvm/vm.py
```diff
@@ -208,6 +252,100 @@ def shutdown(self, check_vm_up_on_transaction=True, transaction=None):
...
+            for retry in range(timeout):
+                log.info('Waiting for VM "{}" to shutdown'.format(
+                    self.dataset_obj['hostname']))
+                vm_status_code = self.aws_describe_instance_status(
+                    self.dataset_obj['aws_instance_id'])
+                if vm_status_code == AWS_RETURN_CODES['stopped']:
+                    log.info('"{}" is stopped.'.format(
+                        self.dataset_obj['hostname']))
+                    break
+                time.sleep(1)
```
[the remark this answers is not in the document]
lacunoc: This is much simpler and working as needed. Thanks!

igvm/vm.py
```diff
@@ -335,6 +472,95 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
 
         log.info('"{}" is successfully built.'.format(self.fqdn))
 
+    def aws_build(self,
+                  run_puppet: bool = True,
+                  debug_puppet: bool = False,
+                  postboot: str = None,
```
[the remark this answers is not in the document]
lacunoc: changed to `Optional[str]`

igvm/vm.py
```diff
@@ -335,6 +472,95 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
...
+                    InstanceIds=[self.dataset_obj['aws_instance_id']],
+                    WaiterConfig={
+                        'Delay': 15,
+                        'MaxAttempts': int(timeout_cloud_init / 15),
+                    }
+                )
+                cloud_init.update(int(timeout_cloud_init - (
+                    int(timeout_cloud_init / 4))))
+                break
+            time.sleep(1)
```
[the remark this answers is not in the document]
lacunoc: I will leave it like it is for now.

igvm/vm.py
```diff
@@ -423,38 +647,38 @@ def run_puppet(self, clear_cert=False, debug=False):
 
         if clear_cert:
             with settings(
-                host_string=self.dataset_obj['puppet_ca'],
-                user='root',
-                warn_only=True,
+                    host_string=self.dataset_obj['puppet_ca'],
+                    user='root',
+                    warn_only=True,
```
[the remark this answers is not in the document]
lacunoc: reverted

igvm/vm.py
```diff
@@ -384,7 +610,7 @@ def prepare_vm(self):
         # Copy resolv.conf from Hypervisor
         fd = BytesIO()
         with self.hypervisor.fabric_settings(
-            cd(self.hypervisor.vm_mount_path(self))
+                cd(self.hypervisor.vm_mount_path(self))
```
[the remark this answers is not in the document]
lacunoc: reverted

igvm/vm.py
```diff
@@ -335,6 +472,95 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
 
         log.info('"{}" is successfully built.'.format(self.fqdn))
 
+    def aws_build(self,
+                  run_puppet: bool = True,
+                  debug_puppet: bool = False,
+                  postboot: str = None,
+                  timeout_vm_setup: int = 300,
+                  timeout_cloud_init: int = 600):
```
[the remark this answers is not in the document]
lacunoc: I missed adding the return type, thanks! Now added

igvm/commands.py
```diff
@@ -95,13 +97,17 @@ def vcpu_set(vm_hostname, count, offline=False):
...
+
         _check_defined(vm)
 
         if offline and not vm.is_running():
             log.info(
-                '"{}" is already powered off, ignoring --offline.'
-                .format(vm.fqdn)
-            )
+                '"{}" is already powered off, ignoring --offline.'.format(
+                    vm.fqdn))
```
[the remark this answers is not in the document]
lacunoc: I tried to revert every changed `.format()` and also changed more wrongly used ones.

lacunoc:
retest this please

igvm/vm.py
```diff
@@ -197,6 +201,46 @@ def start(self, force_stop_failed=True, transaction=None):
...
+            if 'DryRunOperation' not in str(e):
+                raise
+
+        try:
+            response = ec2.start_instances(
+                InstanceIds=[self.dataset_obj['aws_instance_id']],
+                DryRun=False
+            )
+            current_state = response['StartingInstances'][0][
+                'CurrentState']['Code']
```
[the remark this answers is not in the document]
lacunoc: Then the line is too long. What's your proposal?

igvm/vm.py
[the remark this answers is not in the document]
lacunoc: Will be tackled by TECH-39406 at a later point.

igvm/kvm.py
```diff
@@ -294,12 +286,13 @@ def migrate_live(source, destination, vm, domain):
...
-                        js['disk_processed'] / (js['disk_total']+1) * 100,
-                        js['disk_processed']/1024/1024,
-                        js['disk_total']/1024/1024,
-                        js['memory_processed'] / (js['memory_total']+1) * 100,
-                        js['memory_processed']/1024/1024,
-                        js['memory_total']/1024/1024,
+                        js['disk_processed'] / (js['disk_total'] + 1) * 100,
+                        js['disk_processed'] / 1024 / 1024,
+                        js['disk_total'] / 1024 / 1024,
+                        js['memory_processed'] /
```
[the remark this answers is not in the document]
lacunoc: Reverted the line splitting

igvm/vm.py
```diff
@@ -335,6 +472,95 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
...
+
+        Build a VM in AWS.
+
+        :param: run_puppet: Run puppet (incl. cert clean) after VM creation
+        :param: debug_puppet: Run puppet in debug mode
+        :param: postboot: cloudinit configuration put as userdata
+        :param: timeout_vm_setup: Timeout value for the VM creation
+        :param: timeout_cloud_init: Timeout value for the cloudinit
+                                    provisioning
+        """
```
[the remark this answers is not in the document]
lacunoc: Added the Exception to the docstring

igvm/commands.py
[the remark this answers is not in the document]
HerrSpace: Is it the hostname set inside the VM? IIRC we set that to the FQDN too for igvm created VMs or do I misremember?

igvm/settings.py
```diff
@@ -162,6 +171,108 @@
...
+    'terminated': 48,
+    'stopping': 64,
+    'stopped': 80,
+}
+
+AWS_CONFIG = [
+    {
+        'puppet': {
+            'ca_addr': '212.53.146.132',
+            'master_addr': '212.53.146.131',
```
HerrSpace: Still not using dns names, why?

igvm/templates/aws_user_data.cfg
```diff
@@ -0,0 +1,37 @@
...
+apt:
+  sources:
+    {{ innogames_stable_name }}:
+      source: "{{ innogames_stable_source }}"
+      key: |
+        {{ innogames_stable_key }}
+    {{ basestretch_stable_name }}:
+      source: "{{ basestretch_stable_source }}"
+      key: |
+        {{ basestretch_stable_key }}
```
HerrSpace: You could also generate the sources list here from a loop over AWS_CONFIG['apt'] to not even hardcode the two repos

igvm/settings.py
lacunoc: Because this is just a temporarily solution using our deprecated puppet infrastructure. The plan is  Furthermore we have to use internal name with external address here. We will get rid of this soonish.

igvm/templates/aws_user_data.cfg
```diff
@@ -0,0 +1,37 @@
...
+package_update: true
+package_upgrade: true
+packages:
+ - puppet-agent
+ - puppet-msgpack
+
+runcmd:
+  - echo "# We need to use external IPs for puppet here" >> /etc/hosts
+  - echo {{ puppet_ca_addr }} {{ puppet_ca }} >> /etc/hosts
+  - echo {{ puppet_master_addr }} {{ puppet_master }} >> /etc/hosts
```
HerrSpace: The use of /etc/hosts was always strongly discouraged, if you'd use dns names, you wouldn't have to do this.

lacunoc: @Woellchen Can you shed some light here again? Or isn't this needed anymore?

igvm/vm.py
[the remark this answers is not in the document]
HerrSpace: Hmm, you could:
- Just keep the line too long, sometimes that's the least bad option.
- Or do this:
```python
            current_state = (
                response['StartingInstances'][0]['CurrentState']['Code']
            )
```
- Or this:
```python
            response = ec2.start_instances(
                InstanceIds=[self.dataset_obj['aws_instance_id']],
                DryRun=False
            )['StartingInstances'][0]
            current_state = response['CurrentState']['Code']
```

Either way this makes a lot of assumptions about the data structure returned by boto. Such large datastructure accesses often highlight code places with too little error handling. It's easy to imagin error cases in which the api returns something weird and you run into an IndexError or KeyError here. I tried to have a look at the boto(core) code, but it's quite abstract so I couldn't come to a good conclusion whether this might be safe after all.

igvm/settings.py
HerrSpace: Ah, really? I thought we already were using a public name with the old vm_externals. Okay then, but maybe add a FIXME here.

igvm/templates/aws_user_data.cfg
Woellchen: in case of external dcs and puppet this is still necessary and won't change in near future because of different reasons i don't really feel like exposing on a public github repo :)

igvm/settings.py
lacunoc: I've added a TODO

igvm/templates/aws_user_data.cfg
lacunoc: I've added a TODO

igvm/templates/aws_user_data.cfg
lacunoc: This is now generated with a loop

igvm/vm.py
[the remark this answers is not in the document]
lacunoc: fixed

igvm/templates/aws_user_data.cfg
HerrSpace: Looking good! <3

lacunoc:
retest this please

igvm/vm.py
```diff
@@ -335,6 +475,106 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
...
+                                    cloud_init.update(
+                                        timeout_cloud_init - retry - 1)
+                                    break
+                            except (SystemExit, NetworkError) as e:
+                                time.sleep(1)
+                        else:
+                            time.sleep(1)
+                break
+            time.sleep(1)
+
```
HerrSpace: May I offer the slightly more readable:

```python3

        # Wait for AWS to decalre the VM running
        while (
            timeout_vm_setup and
            AWS_RETURN_CODES['running'] != self.aws_describe_instance_status(
                self.dataset_obj['aws_instance_id']
            )
        ):
            vm_setup.update(1)
            timeout_vm_setup -= 1
            time.sleep(1)
        vm_setup.update(timeout_vm_setup - 1)
        # TODO: Handle overrun timeout

        # Try to provision the VM
        for retry in range(timeout_cloud_init):
            cloud_init.update(1)

            # Only try to connect every 10s to avoid paramiko exceptions
            if retry % 10 != 0:
                time.sleep(1)
                continue

            with settings(
                hide('aborts'),
                host_string=self.dataset_obj['hostname'],
                warn_only=True,
                abort_on_prompts=True,
            ):
                try:
                    if run(
                        'find /var/lib/cloud/instance/ boot-finished',
                        quiet=True
                    ).succeeded:
                        cloud_init.update(timeout_cloud_init - retry - 1)
                        break
                except (SystemExit, NetworkError):
                    time.sleep(1)
        # TODO: Handle overrun timeout
```

HerrSpace: Saving you four levels of indent.

igvm/commands.py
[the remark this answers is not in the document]
HerrSpace: The closing parenthesis are still the same though..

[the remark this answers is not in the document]
lacunoc: Are you sure you're checking the latest pushed version?
https://github.com/innogames/igvm/pull/120/commits/0bca873d14e42afb5c278518f1e6c29a58c94123#diff-60a07663873f7ef87bb0c4c69ede494fL102-L104

igvm/vm.py
lacunoc: Looks promising, will check it! Thank you

lacunoc: Worked like a charme, added!

igvm/commands.py
```diff
@@ -235,45 +255,74 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
...
                offline=False, offline_transport='drbd',
                allow_reserved_hv=False, no_shutdown=False):
     """Migrate a VM to a new hypervisor."""
+
     with ExitStack() as es:
         vm = es.enter_context(
             _get_vm(vm_hostname, allow_retired=True)
         )
+
+        if vm.dataset_obj['igvm_operation_mode'] != 'kvm':
```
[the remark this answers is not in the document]
lacunoc: Thanks for your extensive review and feedback. We will leave it like is for now, since it's pretty easy to rename this, if we're planning to virtualize reasonable things without igvm. But I noted it

PR #122, opened 2019-10-02
fix: AWS user date template path incomplete

kofrezo:


igvm/commands.py
```diff
@@ -12,7 +12,7 @@
 from fabric.colors import green, red, white, yellow
 from fabric.network import disconnect_all
 from ipaddress import ip_address
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import Environment, FileSystemLoader, PackageLoader
```
lacunoc: The `FileSystemLoader` is not used anymore. We can remove this import.

PR #126, opened 2019-10-21
Make IP address change more reliable

TuxPowered42:


igvm/commands.py
```diff
@@ -224,32 +228,47 @@ def change_address(vm_hostname, new_address, offline=False):
...
+        new_network = Query(
+            {
+                'servertype': 'route_network',
+                'state': 'online',
+                'network_type': 'internal',
+                'intern_ip': Contains(new_address),
+            }
+        ).get()['hostname']
+
+        if vm.dataset_obj['intern_ip'] == new_address:
```
Woellchen: we can save the above query if we make this check above

igvm/commands.py
```diff
@@ -224,32 +228,47 @@ def change_address(vm_hostname, new_address, offline=False):
...
             raise ConfigError('New IP address is the same as the old one!')
 
-        vm_was_running = vm.is_running()
-
-        vm.dataset_obj['intern_ip'] = new_address
-        vm.dataset_obj.commit()
+        if not vm.hypervisor.get_vlan_network(new_address) and not migrate:
```
Woellchen: same here. move the check above the query

igvm/commands.py
```diff
@@ -224,32 +228,47 @@ def change_address(vm_hostname, new_address, offline=False):
...
 
-        if vm_was_running:
-            vm.shutdown()
+        vm_was_running = vm.is_running()
 
-        try:
-            with Transaction() as transaction:
-                vm.hypervisor.mount_vm_storage(vm, transaction)
+        with Transaction() as transaction:
+            if vm_was_running:
```
Woellchen: hm. shouldn't we actually check for `offline == True` additionally?

igvm/commands.py
```diff
@@ -224,32 +228,47 @@ def change_address(vm_hostname, new_address, offline=False):
...
+                vm.shutdown(
+                    transaction=transaction,
+                    check_vm_up_on_transaction=False,
+                )
+            vm.change_address(new_address, new_network, transaction=transaction)
+
+            if migrate:
+                vm_migrate(
+                    vm_object=vm,
+                    run_puppet=True, offline=True, no_shutdown=True,
```
Woellchen: same here. shouldn't we check for `offline == True`? e.g. `offline=offline`? what about `offline_transport` then? can we support drbd when `offline=True` and `offline_transport='drbd'`? maybe we need a few more checks here

igvm/commands.py
```diff
@@ -317,40 +336,52 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
...
+def vm_migrate(vm_hostname=None, vm_object=None, hypervisor_hostname=None,
                run_puppet=False, debug_puppet=False,
                offline=False, offline_transport='drbd',
                allow_reserved_hv=False, no_shutdown=False):
     """Migrate a VM to a new hypervisor."""
 
-    with ExitStack() as es:
-        vm = es.enter_context(
-            _get_vm(vm_hostname, allow_retired=True)
+    if (
```
Woellchen: how about `if not bool(vm_hostname) ^ bool(vm_object)`?

igvm/commands.py
TuxPowered42: Noppe, this whole function can work only offline and this flag is checked in the beginning of it.

igvm/commands.py
TuxPowered42: Noppe, this whole function can work only offline and this flag is checked in the beginning of it.

igvm/cli.py
```diff
@@ -198,6 +198,24 @@ def parse_args():
         action='store_true',
         help='Perform IP address change offline',
     )
+    subparser.add_argument(
```
kofrezo: Why adding migration support there is a dedicated command igvm migrate. I don't why this is beneficial could you please explain.

igvm/commands.py
```diff
@@ -224,32 +228,48 @@ def change_address(vm_hostname, new_address, offline=False):
                     vm.dataset_obj['igvm_operation_mode'])
             )
 
-        old_address = vm.dataset_obj['intern_ip']
         new_address = ip_address(new_address)
 
-        if old_address == new_address:
+        if vm.dataset_obj['intern_ip'] == new_address:
```
kofrezo: What about IPv6 address. Won't it be broken after changing the network ?

igvm/cli.py
TuxPowered42: Because some users demanded functionality of changing IP address of a VM to a subnet not supported on original HV. To implemment it migration is necessary. Other way of doing it would be to have `igvm migrate --change-address` but then there would be two ways of changing address. I would personally be totally fine with *not* providing users such functionality but apparently it's a hot topic and probably the most wanted one.

igvm/commands.py
TuxPowered42: It will be. We could move functionality of `add_ipv6_address` script to Serveradmin_extras to solve this once and for all.

TuxPowered42: I consider this one resolved because it is done as good as possible now, and once Serveradmin auto-assigns IPv6 this issue will be fully resolved.

TuxPowered42:
retest this, please

TuxPowered42:
Tests have passed but test reporting is still broken.

PR #127, opened 2019-10-24
feat(aws): Add disk-set, sync and some cleanup

lacunoc:


igvm/vm.py
```diff
@@ -713,3 +714,75 @@ def unblock_autostart(self):
...
+        ec2 = boto3.client('ec2')
+        ec2.modify_volume(VolumeId=volume_id, Size=int(size))
+
+        time.sleep(5)
+
+        with settings(
+            host_string=self.dataset_obj['hostname'],
+            warn_only=True,
+        ):
+            disk = sudo('lsblk -r | grep disk | cut -d " " -f1')
```
TuxPowered42: Please never use sudo from Fabric, we have `vm.run()` which handles that.

lacunoc: I changed all calls to use `vm.run()`

igvm/vm.py
```diff
@@ -713,3 +714,75 @@ def unblock_autostart(self):
...
+
+        response = ec2.Instance(self.dataset_obj['aws_instance_id'])
+        for vol in response.volumes.all():
+            volume_id = vol.id
+            break
+
+        ec2 = boto3.client('ec2')
+        ec2.modify_volume(VolumeId=volume_id, Size=int(size))
+
+        time.sleep(5)
```
TuxPowered42: This is risky, wouldn't it be better to check disk size in a loop and wait for it to really grow?

lacunoc: Changed to while loop, which is checking the actual disk_size with `lsblk -bdnro size` before trying to increase the filesystem.

igvm/vm.py
```diff
@@ -713,3 +714,75 @@ def unblock_autostart(self):
...
+
+        with settings(
+            host_string=self.dataset_obj['hostname'],
+            warn_only=True,
+        ):
+            disk = self.run('lsblk -r | grep disk | cut -d " " -f1')
```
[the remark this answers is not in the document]
lacunoc: changed

igvm/vm.py
```diff
@@ -713,3 +714,75 @@ def unblock_autostart(self):
...
+            partition = self.run('lsblk -r | grep part | cut -d " " -f1')
```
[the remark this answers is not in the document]
lacunoc: changed

HerrSpace:
Good stuff!

PR #132, opened 2020-02-19
Run parallel tests with pytest-xdist

TuxPowered42:


kofrezo:
Please mind that running the test in parallel has a high risk of two or more VMs allocating the same IP address which leads to a 403 on comitting the object as seen in the failing tests.

PR #136, opened 2020-02-26
Speed up Hypervisor finding procedure

Woellchen:
Dramatically speed up the process by doing an additional HV filtering
round based on real data before HVs are locked in the next round to
yield it (locked).
Reordering of HV checks in order to execute cheap checks first and
expensive ones last. This can potentially reduce a lot of wasted time.

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -78,11 +78,12 @@ def evacuate(hv_hostname, offline=None, dry_run=False):
 
         for vm in hv.dataset_obj['vms']:
             vm_function = vm['function']
+            is_offline_migration = (
+                offline is not None
+                and (offline == [] or vm_function in offline)
+            )
 
-            if (
-                offline is not None and
-                (offline == [] or vm_function in offline)
-            ):
+            if is_offline_migration:
                 if dry_run:
                     log.info('Would migrate {} offline'.format(vm['hostname']))
                 else:
@@ -238,7 +239,8 @@ def change_address(
             raise ConfigError('New IP address is the same as the old one!')
 
         if not vm.hypervisor.get_vlan_network(new_address) and not migrate:
-            raise ConfigError('Current hypervisor does not support new subnet!')
+            err = 'Current hypervisor does not support new subnet!'
+            raise ConfigError(err)
 
         new_network = Query(
             {
@@ -257,7 +259,9 @@ def change_address(
                     transaction=transaction,
                     check_vm_up_on_transaction=False,
                 )
-            vm.change_address(new_address, new_network, transaction=transaction)
+            vm.change_address(
+                new_address, new_network, transaction=transaction,
+            )
 
             if migrate:
                 vm_migrate(
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -4,14 +4,15 @@ Copyright (c) 2018 InnoGames GmbH
 """
 
 import logging
-from os import environ
+from collections import OrderedDict
 from contextlib import contextmanager, ExitStack
+from ipaddress import ip_address
+from os import environ
 
 from adminapi.dataset import Query
 from adminapi.filters import Any, StartsWith, Contains
 from fabric.colors import green, red, white, yellow
 from fabric.network import disconnect_all
-from ipaddress import ip_address
 from jinja2 import Environment, PackageLoader
 from libvirt import libvirtError
 
@@ -33,7 +34,7 @@ from igvm.settings import (
     VM_ATTRIBUTES,
 )
 from igvm.transaction import Transaction
-from igvm.utils import parse_size
+from igvm.utils import parse_size, parallel
 from igvm.vm import VM
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


igvm/hypervisor.py
```diff
@@ -215,6 +190,40 @@ def check_vm(self, vm, offline):
                     .format(*hw_pair)
                 )
 
+        # HV in supported state?
+        if self.dataset_obj['state'] not in ['online', 'online_reserved']:
```
kofrezo: I think this decision should be done when filtering for the hypervisors to check in the first place. This would make it impossible to change the hypervisor selection from the commandline with arguments such as --ignore-reserved.

See: https://github.com/innogames/igvm/blob/master/igvm/commands.py#L312

igvm/hypervisor.py
```diff
@@ -215,6 +190,40 @@ def check_vm(self, vm, offline):
...
+            raise InvalidStateError(
+                'Hypervisor "{}" is not in online state ({}).'
+                .format(self.fqdn, self.dataset_obj['state'])
+            )
+
+        # Enough CPUs?
+        if vm.dataset_obj['num_cpu'] > self.dataset_obj['num_cpu']:
```
kofrezo: I think we allow CPU over allocation no ?

igvm/hypervisor.py
```diff
@@ -215,6 +190,40 @@ def check_vm(self, vm, offline):
...
+            raise HypervisorError(
+                'Not enough CPUs. Destination Hypervisor has {0}, '
+                'but VM requires {1}.'
+                .format(self.dataset_obj['num_cpu'], vm.dataset_obj['num_cpu'])
+            )
+
+        # Proper VLAN?
+        if not self.get_vlan_network(vm.dataset_obj['intern_ip']):
```
kofrezo: This check also has taken place already see: 

https://github.com/innogames/igvm/blob/fdc85aa8825a07abf24448df0b7081f817d91f80/igvm/commands.py#L833

igvm/hypervisor.py
```diff
@@ -215,6 +190,40 @@ def check_vm(self, vm, offline):
...
+                'Hypervisor "{}" does not support route_network "{}".'
+                .format(self.fqdn, vm.route_network)
+            )
+
+        # Those checks below all require libvirt connection,
+        # so execute them last to avoid unnecessary overhead if possible.
+
+        # Enough memory?
+        free_mib = self.free_vm_memory()
+        if vm.dataset_obj['memory'] > free_mib:
```
kofrezo: This one is arguable. I think it is safe for us but maybe this should still remain a hypervisor preference because people might want to allow memory ballooning. What is your thought on this @TuxPowered42

igvm/commands.py
```diff
@@ -4,14 +4,19 @@
...
+from collections import OrderedDict
 from contextlib import contextmanager, ExitStack
+from ipaddress import ip_address
+from os import environ
 
 from adminapi.dataset import Query
 from adminapi.filters import Any, StartsWith, Contains
 from fabric.colors import green, red, white, yellow
+from fabric.context_managers import settings
+from fabric.decorators import parallel
```
TuxPowered42: Any chance of doing this via standard Python methods? I would prefer to keep Fabric dependencies at minimum.

igvm/commands.py
```diff
@@ -78,10 +83,12 @@ def evacuate(hv_hostname, offline=None, dry_run=False):
...
-            if (
-                offline is not None and
-                (offline == [] or vm_function in offline)
-            ):
+            is_offline_migration = (
+                offline is not None
+                and (offline == [] or vm_function in offline)
+            )
+
+            if is_offline_migration:
```
TuxPowered42: Please split cosmetic changes into a separate commit.

igvm/commands.py
```diff
@@ -827,57 +838,60 @@ def _get_hypervisor(hostname, allow_reserved=False):
...
+    for possible_hv in possible_hvs.values():
         try:
-            hypervisor.check_vm(vm, offline)
-        except libvirtError as error:
-            hypervisor.release_lock()
-            log.warning(
-                'Preferred hypervisor "{}" is skipped: {}'.format(
-                    hypervisor, error)
-            )
+            possible_hv.acquire_lock()
```
kofrezo: Will this lock be release if an exception is thrown ?

igvm/commands.py
```diff
@@ -827,57 +838,60 @@ def _get_hypervisor(hostname, allow_reserved=False):
...
-            )
+
+        if not _check_vm(possible_hv, vm, offline):
+            possible_hv.release_lock()
             continue
 
         try:
-            yield hypervisor
+            yield possible_hv
+            return
```
kofrezo: yield return makes no sense. You can just return.

igvm/commands.py
Woellchen: good catch. fill fix

Woellchen: actually when this throws an exception there is in fact no lock. so nothing to unlock

igvm/commands.py
Woellchen: true story

igvm/hypervisor.py
TuxPowered42: Interesting observation. But this code is then even more wrong: a HV could change its state whille it's being searched for. We should check here for exactly requested state.

igvm/hypervisor.py
TuxPowered42: Noppe.

igvm/commands.py
Woellchen: yeah you can probably fiddle with `multiprocessing` or similar.. but no please not :D fabric makes it real easy in this case to parallelize

igvm/hypervisor.py
TuxPowered42: Same as with state, we check a HV for all parameters while its locked. However Serveradmin users could change this without locking.

If we don't care for that, then this is redundant code, yes.

igvm/hypervisor.py
Woellchen: lets wait for feedback from @TuxPowered42 but addresse this in another pr since the logic is the same as before, just in another place

igvm/hypervisor.py
Woellchen: lets wait for feedback from @TuxPowered42 but addresse this in another pr since the logic is the same as before, just in another place

igvm/hypervisor.py
Woellchen: lets wait for feedback from @TuxPowered42 but addresse this in another pr since the logic is the same as before, just in another place

igvm/hypervisor.py
Woellchen: lets wait for feedback from @TuxPowered42 but addresse this in another pr since the logic is the same as before, just in another place

TuxPowered42: We don't allow for memory ballooning. We under-allocate VMs and give them their own swap.

igvm/commands.py
TuxPowered42: With the style change of "for ... else: throw" to "for ... return; throw" this seems necessary. Why the style change in the first place?

igvm/hypervisor.py
TuxPowered42: If it's redundant we could make a separate PR to fix it. I would prefer not to change it too much with other changes @Woellchen did.

igvm/commands.py
Woellchen: actually i agree to @kofrezo. this is not a generator. it returns exactly one hypervisor. i will remove the yield statement

Woellchen: i just reiterated on this. `yield` is necessary because this function is used as a context manager in order to make sure hv is always locked and unlocked correctly. i don't see a point to use an `else` statement. when the outer code is done working on the hv, this context manager is done, hence return

Woellchen: i reverted to the `for/else` construct

igvm/commands.py
Woellchen: ok after discussion i have implemented a simple parallel executor based on threads without the use of fabric. this also prevents libvirt connections from being established more than once since fabric's `parallel` is based on subprocesses

Woellchen:
ready when you are

PR #137, opened 2020-03-03
Refactor so that clean cert is not called on igvm build

eviscares:


igvm/vm.py
```diff
@@ -457,6 +457,10 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
 
             if run_puppet:
                 self.run_puppet(clear_cert=True, debug=debug_puppet)
+                if rebuild:
+                    self.run_puppet(clear_cert=True, debug=debug_puppet)
```
Woellchen: how about `self.run_puppet(clear_cert=rebuild, debug=debug_puppet)`

igvm/vm.py
```diff
@@ -457,6 +457,10 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
 
             if run_puppet:
                 self.run_puppet(clear_cert=True, debug=debug_puppet)
```
Woellchen: we are running puppet twice now

igvm/commands.py
```diff
@@ -325,11 +325,15 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
 
             if rebuild and vm.hypervisor.vm_defined(vm):
                 vm.hypervisor.undefine_vm(vm)
+                is_rebuild = True
```
Woellchen: how about `is_rebuild = rebuild and vm.hypervisor.vm_defined(vm)`
and then do the if based on `is_rebuild` directly. saves us the `else`

igvm/vm.py
eviscares: Shiny, I like it.

igvm/vm.py
eviscares: Got it, removed

igvm/commands.py
eviscares: Changed

TuxPowered42:
Retest this, please.

igvm/commands.py
```diff
@@ -328,13 +328,14 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
                     '"{}" is still running.'.format(vm.fqdn)
                 )
 
-            if rebuild and vm.hypervisor.vm_defined(vm):
+            if vm.hypervisor.vm_defined(vm):
```
TuxPowered42: This invalidates the idea of "rebuild" being a different operation than "build". If somebody has a stopped VM, it will be undefined without asking and then building will fail because it's disk is still present.

igvm/vm.py
```diff
@@ -430,7 +430,7 @@ def info(self):
             result['status'] = 'new'
         return result
 
-    def build(self, run_puppet=True, debug_puppet=False, postboot=None):
+    def build(self, run_puppet=True, debug_puppet=False, postboot=None, rebuild=False):
```
TuxPowered42: Why do you call this `rebuild` when it only controls clearing of Puppet certificate?

tests/test_integration.py
```diff
@@ -378,7 +378,7 @@ def test_rebuild(self):
 class CommandTest(IGVMTest):
     def setUp(self):
         super(CommandTest, self).setUp()
-        vm_build(VM_HOSTNAME)
+        vm_build(VM_HOSTNAME, rebuild=True)
```
TuxPowered42: I'm surprised this was not necessary on BuildTest too. Did you get lucky just because of order of tests?

igvm/vm.py
```diff
@@ -456,7 +456,7 @@ def build(self, run_puppet=True, debug_puppet=False, postboot=None):
             self.prepare_vm()
 
             if run_puppet:
-                self.run_puppet(clear_cert=True, debug=debug_puppet)
+                self.run_puppet(clear_cert=rebuild, debug=debug_puppet)
```
TuxPowered42: If the new solution is to clean Puppet cert when a VM gets deleted, why bother clearing it on build at all? Maybe all you need to do is to call cert clearning from tests and force users to delete VMs correctly during normal operation?

tests/test_integration.py
eviscares: If so, we got lucky every single time. Also some of them have rebuild set to true.

igvm/vm.py
eviscares: Because it is only required in the rebuild case.

igvm/vm.py
eviscares: It was always the concept to clean on deletion. The old behaviour to always clean on build was what gave us this issue initially. The only situation during which we want to clean certs is when we rebuild.

igvm/commands.py
eviscares: That is reasonable. I'll change that back.

igvm/vm.py
TuxPowered42: Flags' names should tell us what they do, not why we call them.

igvm/vm.py
TuxPowered42: @eviscares wrote:

> It was always the concept to clean on deletion.

The diff says there was no clearing on deletion before https://github.com/innogames/igvm/pull/137/files#diff-60a07663873f7ef87bb0c4c69ede494fR561

tests/test_integration.py
TuxPowered42: It failed again on a random test. The only one with `rebuid=True` is `test_rebuild()`. We should not write code depending on particular order of tests and as far as I understand the 1st test can't be run as rebuild because it will not be able to undefine the VM.

igvm/vm.py
eviscares: Yes, and that is what my proposed change is about, to clean on deletion, instead of on build, because it makes more sense that way.

igvm/vm.py
eviscares: Renamed the flag to 'clean_cert'

igvm/vm.py
TuxPowered42: Your change adds conditional cleaning on build and on unconditional on deletion. Maybe we could get away without the former one at all.

igvm/utils.py
```diff
@@ -249,4 +249,4 @@ def parallel(
         else:
             results.append(result)
 
-    return results
+    return results
```
TuxPowered42: Newline missing.

igvm/commands.py
```diff
@@ -329,12 +330,15 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
                 )
 
             if rebuild and vm.hypervisor.vm_defined(vm):
+                log.warning("Undefining VM")
```
TuxPowered42: Looks like debug leftover.

igvm/utils.py
```diff
@@ -8,14 +8,14 @@
 import logging
 import socket
 import time
+import random
```
TuxPowered42: Not used anymore?

tests/test_integration.py
```diff
@@ -580,7 +583,7 @@ def test_info(self):
 class MigrationTest(IGVMTest):
     def setUp(self):
         super(MigrationTest, self).setUp()
-        vm_build(VM_HOSTNAME)
+        vm_build(VM_HOSTNAME, rebuild=True)
```
TuxPowered42: Is this still needed since we clean cert before each test?

tests/test_integration.py
TuxPowered42: Is this still needed since we clean cert before each test?

tests/test_integration.py
eviscares: Probably not. Will remove

igvm/utils.py
eviscares: Added

igvm/commands.py
eviscares: Removed

igvm/utils.py
eviscares: Removed

PR #139, opened 2020-03-09
Fix typo on ca_host variable name

lamaral:
There was an extra S in the ca_host variable name that was breaking the `build --rebuild` command for VMs that have a loadbalancer as ca_host instead of a VM.


lamaral:
Retest, please

PR #142, opened 2020-03-11
Rework the VM renaming function

lamaral:
The changes in 778907e broke the VM renaming, as the operations now
rely on hostname and we were updating it to the new one before shutting
down and redefining the VM.

Puppet actually already takes care of putting data into `/etc/hosts`, `/etc/hostname` and `/etc/mailname`, so we can actually do a puppetrun to get everything setup.

### DIFF
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -634,6 +634,7 @@ class VM(Host):
             self.shutdown(transaction=transaction)
 
             self.hypervisor.redefine_vm(self, new_fqdn=new_hostname)
+            print(self.previous_hostname)
             transaction.on_rollback(
                 'undo_redefine',
                 self.hypervisor.redefine_vm,
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -160,6 +160,23 @@ class VM(Host):
         """Resizes the host memory."""
         self.hypervisor.vm_set_memory(self, memory)
 
+    def set_hostname(self, new_hostname, transaction=None):
+        """Changes the hostname of a VM"""
+        self.previous_hostname = self.dataset_obj['hostname']
+
+        self.dataset_obj['hostname'] = new_hostname
+        self.check_serveradmin_config()
+
+        self.dataset_obj.commit()
+
+        if transaction:
+            transaction.on_rollback('reset_hostname', self.revert_hostname)
+
+    def revert_hostname(self):
+        """Revert the VM name to the previously defined name"""
+        if hasattr(self, 'previous_hostname'):
+            self.set_hostname(self.previous_hostname)
+
     def check_serveradmin_config(self):
         """Validate relevant Serveradmin attributes"""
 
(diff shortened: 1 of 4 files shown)


igvm/vm.py
```diff
@@ -588,29 +605,26 @@ def aws_build(self,
...
+                'undo_redefine',
+                self.hypervisor.redefine_vm,
+                self,
+                new_fqdn=self.previous_hostname
+            )
 
-            self.dataset_obj.commit()
+            self.hypervisor.mount_vm_storage(self, transaction=transaction)
+
+            time.sleep(5)
```
Woellchen: what do we need this `sleep` for now?

igvm/vm.py
```diff
@@ -151,6 +151,23 @@ def set_memory(self, memory):
...
+        """Changes the hostname of a VM"""
+        self.previous_hostname = self.dataset_obj['hostname']
+
+        self.dataset_obj['hostname'] = new_hostname
+        self.check_serveradmin_config()
+
+        self.dataset_obj.commit()
+
+        if transaction:
+            transaction.on_rollback('reset_hostname', self.revert_hostname)
```
Woellchen: lets be consistent and dont mix names. decide for either `reset_hostname` or `revert_hostname`

igvm/vm.py
```diff
@@ -151,6 +151,23 @@ def set_memory(self, memory):
         """Resizes the host memory."""
         self.hypervisor.vm_set_memory(self, memory)
 
+    def set_hostname(self, new_hostname, transaction=None):
+        """Changes the hostname of a VM"""
+        self.previous_hostname = self.dataset_obj['hostname']
```
Woellchen: i really don't like this. please try to pass this information in the `on_rollback` callback already. you can do something like `transaction.on_rollback('reset_hostname', lambda: self.revert_hostname(self.dataset_obj['hostname']))`. obviousely before you change it in the `dataset_obj`. or save it to `previous_hostname` at least. but there is no reason to make this an instance variable is there? just my opinion

Woellchen: or rather don't use lambdas and use the `args`/`kwargs` of `on_rollback`

igvm/vm.py
lamaral: Serveradmin takes some time to commit the data. 

If we start the puppetrun straight away, the Puppet ENC will not see the new hostname there.

igvm/vm.py
lamaral: Oops, this was some leftover.

igvm/vm.py
lamaral: I did this way to try to be consistent with what `reset_state` was already doing. 

I can move it back into the `rename` function also

igvm/vm.py
Woellchen: but as far as i understand the `commit` call is blocking. is this really a problem?

igvm/vm.py
Woellchen: ah i see. then i am fine with it for now. this is not a blocker

igvm/vm.py
lamaral: I'll retest, but afair, yesterday it caused me problems.

lamaral:
Restest this, please

lamaral:
I tested this extensively and the rename works fine. 

The revert operation works half fine, since part of the `redefine_vm()` relies on `vm.dataset_obj['hostname']` and other parts on `vm.fqdn`. The end result is that the revert of the redefine keeps the new hostname. 

I will now write some tests for it, since the renaming isn't being tested at all.

lamaral:
Retest this, please

lamaral:
Retest this please

igvm/vm.py
kofrezo: @Woellchen is right. Committing to Serveradmin with adminapi is a blocking operation. There is no way the function returns before the data is safely committed to the database unless you run it in a synchronous way manually.

kofrezo:
Retest this, please

kofrezo:
Retest this, please

PR #144, opened 2020-03-18
Drop support for vm and lv names not containing object_id

HerrSpace:
This change makes matching libvirt domains and logical volumes to
serveradmin objects safer, by always matching them by their object_id.
The object_id is immutable while a hostname might change.

I introduced the object_id naming almost two years ago so assuming all
VMs were offline migrated or igvm restarted at least once in that time
frame, the legacy name support is no longer required.

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -88,15 +88,9 @@ class Hypervisor(Host):
 
     def get_volume_by_vm(self, vm):
         """Get logical volume information of a VM"""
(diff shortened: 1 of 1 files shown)


Woellchen:
thanks for your patch! i will check today whether this is really the case or if we must still fix a few domain names before

Woellchen:
hold on, almost there 😄

HerrSpace:
image

Woellchen:
thank you @HerrSpace :)

PR #145, opened 2020-03-27
Fix testing setup

Woellchen:
this overhauls the whole testing setup a bit in a way that it's multiprocessing safe.
also it adds more robust puppet cert cleaning and testing vm/hv cleanups.

### DIFF
--- a/igvm/puppet.py
+++ b/igvm/puppet.py
@@ -65,19 +65,14 @@ def clean_cert(vm, user=None, retries=60):
         version = sudo('/usr/bin/puppet --version', shell=False, quiet=True)
 
         if not version.succeeded or int(version.split('.')[0]) < 6:
-            res = sudo(
-                '/usr/bin/puppet cert verify {} '
-                '&& /usr/bin/puppet cert clean {}'.format(
-                    vm['hostname'], vm['hostname'],
-                ),
-            )
-
-            retry = 0
-            while res.return_code == 3 and retry < retries:
-                sleep(1)
-                retry += 1
+            # Every signing and revoking will have the CA regenerate the CRL
+            # file. There are already known problems in Puppet with dealing
+            # with such CRLs. Now if we revoke and/or sign some certificates
+            # in parallel, there is a chance we receive an OpenSSL error (3).
+            # In that case we cannot do anything but retry the operation.
+            for retry in range(1, retries + 1):
                 logger.info(
-                    'Retrying to clear certificate (retry {}/{})'.format(
+                    'Trying to clear certificate ({}/{})'.format(
                         retry, retries,
                     )
                 )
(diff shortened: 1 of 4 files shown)


tests/test_integration.py
```diff
@@ -32,262 +27,107 @@
...
+        self.vm_obj['hypervisor'] = None
+        self.vm_obj['intern_ip'] = get_next_address(VM_NET, 1)
+        self.vm_obj['memory'] = 2048
+        self.vm_obj['no_monitoring'] = True
+        self.vm_obj['num_cpu'] = 2
+        self.vm_obj['os'] = 'stretch'
+        self.vm_obj['project'] = 'test'
+        self.vm_obj['puppet_environment'] = None
+        self.vm_obj['repositories'] = [
             'int:basestretch:stable',
```
lamaral: Isn't it better to start testing with basebuster?

tests/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+    if vm_hostname is None:
+        pattern = '^([0-9]+_)?{}$'.format(
+            VM_HOSTNAME_PATTERN.format(JENKINS_EXECUTOR, '[0-9]+'),
+        )
+    else:
+        pattern = '^([0-9]+_)?{}$'.format(vm_hostname)
+
+    # Clean HVs one by one.
+    for hv in hvs:
+        clean_hv(hv, pattern)
```
lamaral: If two people run manual tests at the same time, wouldn't one cleanup the other, since the executor would be the same (manual) and the pattern is matching any numbers?

lamaral: Ok, thinking again, the idea is not to have two people running tests at once, but just parallelizing one testrun, right?

tests/test_integration.py
Woellchen: i was also surprised. but lets change this in another pr

tests/conftest.py
Woellchen: it was never possible to run more than one manual test at the same time. so this changes nothing in that regard. but yes they would clean up each other if that happens :D but it would also have happened before.
but yes it is fully intentional that tests clean up everything that was left

igvm/puppet.py
```diff
@@ -62,9 +65,29 @@ def clean_cert(vm, user=None):
...
+                        retry, retries,
+                    )
+                )
+                res = run(
+                    'sudo /usr/bin/puppet cert verify {} '
+                    '&& sudo /usr/bin/puppet cert clean {}'.format(
+                        vm['hostname'], vm['hostname'],
+                    ),
+                    shell=False,
+                )
```
lamaral: How about:
```python
            for retry in range(0, retries):
                logger.info(
                    'Trying to clear certificate (retry {}/{})'.format(
                        retry+1, retries,
                    )
                )
                res = run(
                    'sudo /usr/bin/puppet cert verify {} '
                    '&& sudo /usr/bin/puppet cert clean {}'.format(
                        vm['hostname'], vm['hostname'],
                    ),
                    shell=False,
                )
                if res.return_code != 3:
                    break
```
This would remove the duplication of the puppet command.

Woellchen: thank you :). i have to admit my solution is very ugly. yours is much better. will add it quickly

PR #149, opened 2020-05-08
feat(migration): Only migrate if predicted CPU usage is below threshold

lacunoc:
The calculated performance_value of the VM is used to predict
the CPU usage done by the VM on the destination hypervisor.
If the hypervisor then gets above the loadtested threshold,
which would lead to steal time, we dismiss this HV as a candidate.

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -9,6 +9,7 @@ from contextlib import contextmanager
 from time import sleep
 from xml.etree import ElementTree
 
+from igvm.vm import VM
 from libvirt import VIR_DOMAIN_SHUTOFF
 
 from igvm.drbd import DRBD
@@ -817,3 +818,57 @@ class Hypervisor(Host):
             '| /bin/nc.openbsd -q 1 {2} {3}'
             .format(device, size, *listener)
         )
+
+    def hv_pcpus_used_by_vm(self, vm: VM) -> float:
+        """pCPUs used by VM
+
+        Calculate the physical CPUs the VM will use on a potential destination
+        hypervisor.
+
+        :param: vm: VM object
+
+        :return: Amount of pCPUs as float
+        """
+
+        vm_performance_value = vm.vm_performance_value()
+
+        # We can't save floats in graphite cache of serveradmin and therefore
+        # divide it by the value we multiplied it before.
+        hv_perffactor_dest = self.dataset_obj['cpu_perffactor'] / 1000
+        hv_pcpus_dest = vm_performance_value / hv_perffactor_dest
+
+        return float(hv_pcpus_dest)
+
+    def hv_predict_vm_cpu_util(self, vm: VM) -> float:
+        """Predict VM CPU util
+
+        Predict the CPU usage the VM will produce on a potential destination
+        hypervisor.
+
+        :param: vm: VM object
+
+        :return: Cpu utilisation in percent as float
+        """
+
+        hv_pcpus_dest = self.hv_pcpus_used_by_vm(vm)
+        hv_num_cpu_dest = self.dataset_obj['num_cpu']
+        hv_vm_cpu_util_dest = (hv_pcpus_dest / hv_num_cpu_dest) * 100
+
+        return float(hv_vm_cpu_util_dest)
+
+    def hv_cpu_util_overall(self, vm: VM) -> float:
+        """HV CPU util overall
+
+        Calculate the overall CPU usage of a hypervisor incl. the actual
+        migrated VM
+
+        :param: vm: VM object
+
+        :return: Cpu utilisation in percent as float
+        """
+
+        hv_cpu_util_dest = self.hv_predict_vm_cpu_util(vm)
+        hv_cpu_util_pct_dest = self.dataset_obj['cpu_util_pct']
+        hv_cpu_util_overall = hv_cpu_util_pct_dest + hv_cpu_util_dest
+
+        return float(hv_cpu_util_overall)
--- a/igvm/hypervisor_preferences.py
+++ b/igvm/hypervisor_preferences.py
@@ -130,6 +130,36 @@ class HypervisorAttributeValueLimit(object):
         return value is not None and value > self.limit
 
 
+class HypervisorCpuUsageLimit():
+    """Check for CPU usage of the hypervisor incl. the predicted CPU usage
+    of the VM to be migrated.
+
+    Make any hypervisor less likely chosen, which would be above its threshold.
+    """
+    def __init__(self, hardware_model: str, hv_cpu_thresholds: dict):
+        self.hardware_model = hardware_model
+        self.hv_cpu_thresholds = hv_cpu_thresholds
+
+    def __repr__(self):
+        args = repr(self.hardware_model) + ', ' + repr(self.hv_cpu_thresholds)
+
+        return '{}({})'.format(type(self).__name__, args)
+
+    def __call__(self, vm, hv) -> bool:
+        # New VM has no hypervisor attribute yet.
+        if not vm.hypervisor:
+            return False
+
+        hv_model = hv.dataset_obj[self.hardware_model]
+        hv_cpu_threshold = float(self.hv_cpu_thresholds[hv_model])
+        hv_cpu_util_overall = hv.hv_cpu_util_overall(vm)
+
+        return (
+                hv_cpu_util_overall is not None
+                and hv_cpu_util_overall > hv_cpu_threshold
+        )
+
+
 class OverAllocation(object):
     """Check for an attribute being over allocated than the current one"""
     def __init__(self, attribute):
--- a/igvm/settings.py
+++ b/igvm/settings.py
@@ -15,6 +15,7 @@ from igvm.hypervisor_preferences import (
     HashDifference,
     HypervisorAttributeValue,
     HypervisorAttributeValueLimit,
+    HypervisorCpuUsageLimit,
     InsufficientResource,
     OtherVMs,
     OverAllocation,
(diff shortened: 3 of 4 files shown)


lacunoc:
retest this please

igvm/hypervisor.py
```diff
@@ -200,6 +201,32 @@ def check_vm(self, vm, offline):
                 .format(self.dataset_obj['num_cpu'], vm.dataset_obj['num_cpu'])
             )
 
+        # CPU usage of HV below threshold?
+        if vm.hypervisor:
+            vm_load_99 = vm.dataset_obj['load_99'] / 1000
```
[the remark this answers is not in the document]
lacunoc: I've created some helper functions to do the actual calculations and only call the function to get the overall cpu usage for the HV incl. the predicted VM cpu usage. This makes it much more clean and we don't duplicate code anymore.

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
The tests passed btw
https://sysadmin-jenkins.innogames.de/job/integration/job/igvm-testing-github/802/testReport/
Job just wasn't able to send the success message.

igvm/settings.py
```diff
@@ -290,6 +293,17 @@
...
 
+# The loadtested CPU usage thresholds per hypervisor hardware model,
+# before we are experiencing reasonable steal time.
+HYPERVISOR_CPU_THRESHOLDS = {
+    'Dell_M610': 50,  # untested
+    'Dell_M710': 50,  # untested
+    'Dell_M620': 60,
+    'Dell_M630': 70,
+    'Dell_M640': 75,
+    'Dell_R6515': 50,
```
TuxPowered42, line 304: This seems rather low for a brand new hardware with a single NUMA node.

igvm/hypervisor.py
```diff
@@ -200,6 +201,20 @@ def check_vm(self, vm, offline):
...
             )
 
+        # CPU usage of HV below threshold?
+        if vm.hypervisor:
+            hv_cpu_threshold = HYPERVISOR_CPU_THRESHOLDS[
+                self.dataset_obj['hardware_model']]
+            hv_cpu_util_overall = self.hv_cpu_util_overall(vm)
+
+            if hv_cpu_util_overall > hv_cpu_threshold:
+                raise HypervisorError(
```
TuxPowered42, line 211: This part of code is for incompatibility of a HV and a VM. It means that it might be impossible to create new VMs at all. And it won't work nicely with scenarios with HVs with low, occasional load. Maybe it is enough to use this information only for sorting HVs, not to fully exclude them?

lacunoc: Sorting will take place in the preferences already. This shouldn't break initial VM setup, because `vm.hypervisor` is not set.

igvm/settings.py
lacunoc: Well, we loadtested the R6515 with overbooking (96), all CPUs incl. HT (64) and only pCPUs (32) and noticed reasonable stealtime at 50% of CPU load of the HV already. Same we did for the M6X0's, so the values are comparable.

Seems like AMDs HT is not capable of handling our workload properly.

lacunoc:
retest this please

igvm/hypervisor.py
kofrezo: I agree with @TuxPowered42 This should be part of finding a suitable hypervisor in the preferences. The checks in here should just be a safety need to avoid trying to do something that will fail for sure. Creating/Migrating a VM to a loaded hypervisor is probably a bad idea but should not be denied.

igvm/hypervisor.py
```diff
@@ -817,3 +832,57 @@ def device_to_netcat(self, device, size, listener):
             '| /bin/nc.openbsd -q 1 {2} {3}'
             .format(device, size, *listener)
         )
+
+    def hv_pcpus_used_by_vm(self, vm) -> float:
```
kofrezo, line 836: If we support type hints already you can also use some for the parameters (You may want to check the following methods as well).

igvm/hypervisor.py
```diff
@@ -817,3 +832,57 @@ def device_to_netcat(self, device, size, listener):
...
+        hypervisor.
+
+        :param: vm: VM object
+
+        :return: Amount of pCPUs as float
+        """
+
+        vm_performance_value = vm.vm_performance_value()
+
+        hv_perffactor_dest = self.dataset_obj['cpu_perffactor'] / 1000
```
kofrezo, line 849: Better comment here that we can right now not save floats in graphite cache of Serveradmin an therefore divide the value by thousand because we multiplied it in advance.

igvm/hypervisor.py
```diff
@@ -817,3 +832,57 @@ def device_to_netcat(self, device, size, listener):
...
+        hv_pcpus_dest = vm_performance_value / hv_perffactor_dest
+
+        return hv_pcpus_dest
```
kofrezo, line 852: You are not explicit casting to a float. The result of the division might return a integer in case the numbers are even like e.g. 4 / 2 = 2 is this a problem ? (You may want to check the following methods as well)

igvm/hypervisor_preferences.py
```diff
@@ -130,6 +130,36 @@ def __call__(self, vm, hv):
         return value is not None and value > self.limit
 
 
+class HypervisorCpuUsageLimit(object):
```
kofrezo, line 133: Info: The inheritance of object is implicit at least when using Python 3 you don't need to write it.

igvm/hypervisor_preferences.py
```diff
@@ -130,6 +130,36 @@ def __call__(self, vm, hv):
...
+    """Check for CPU usage of the hypervisor incl. the predicted CPU usage
+    of the VM to be migrated.
+
+    Dismiss any hypervisor, which would be above its threshold.
```
kofrezo, line 137: We cannot really dismiss the hypervisor but just make it least likely to be choosen.

igvm/hypervisor_preferences.py
```diff
@@ -130,6 +130,36 @@ def __call__(self, vm, hv):
...
+    """
+    def __init__(self, hardware_model, hv_cpu_thresholds):
+        self.hardware_model = hardware_model
+        self.hv_cpu_thresholds = hv_cpu_thresholds
+
+    def __repr__(self):
+        args = repr(self.hardware_model) + ', ' + repr(self.hv_cpu_thresholds)
+
+        return '{}({})'.format(type(self).__name__, args)
```
kofrezo, line 146: Maybe stick to either format or string concatenation ?

igvm/hypervisor_preferences.py
```diff
@@ -130,6 +130,36 @@ def __call__(self, vm, hv):
...
+        # New VM has no hypervisor attribute yet.
+        if not vm.hypervisor:
+            return False
+
+        hv_model = hv.dataset_obj[self.hardware_model]
+        hv_cpu_threshold = self.hv_cpu_thresholds[hv_model]
+        hv_cpu_util_overall = hv.hv_cpu_util_overall(vm)
+
+        return (
+                hv_cpu_util_overall is not None
```
kofrezo, line 158: What happens if hv_cpu_util_overall is None ? The return value of the method would be None because none of the return statements terminated the method. What happens in such as case should we maybe return True or False instead ?

lacunoc: If `hv_cpu_util_overall` is None, we're automatically returning `False` in every condition
```
>>> a = (None is not None and 78.5 > 70.0)
>>> a
False
>>> a = (62.5 is not None and 62.5 > 70.0)
>>> a
False
>>> a = (78.5 is not None and 78.5 > 70.0)
>>> a
True
```

Only `True` would put the HV to the end of the list.

igvm/hypervisor.py
lacunoc: I removed this part of code, so only the Preference is left.

igvm/hypervisor.py
lacunoc: Added missing type hints.

igvm/hypervisor.py
lacunoc: Comment added

igvm/hypervisor.py
lacunoc: Now everything is casted as `float` explicitly.

igvm/hypervisor_preferences.py
lacunoc: Removed the not needed inheritance.

igvm/hypervisor_preferences.py
lacunoc: Changed the docstring.

igvm/hypervisor_preferences.py
lacunoc: I would like to leave it like it is, since all our Preferences are looking the same.

igvm/hypervisor_preferences.py
kofrezo: Ah true I was confused.

igvm/settings.py
TuxPowered42: If that's the number found by experiment, then I trust it.

igvm/vm.py
```diff
@@ -826,3 +826,21 @@ def aws_sync(self) -> dict:
...
+
+        Calculate the performance_value of the VM by multiplying the
+        load_99 of the VM with the cpu_perffactor of the HV.
+
+        :return: performance_value of VM as float
+        """
+
+        # We can't save floats in graphite cache of serveradmin and therefore
+        # divide it by the value we multiplied it before.
+        vm_load_99 = self.dataset_obj['load_99'] / 1000
```
lacunoc, line 841: We need to use `num_cpu` of the VM instead of the `load_99` value, if ```load_99 > num_cpu```

lacunoc: Added this verification

PR #151, opened 2020-05-20
Prevent crash on unsupported hardware_model

meikomeik:
Prevent crash on unsupported hardware.

### DIFF
--- a/igvm/hypervisor_preferences.py
+++ b/igvm/hypervisor_preferences.py
@@ -25,7 +25,7 @@ class InsufficientResource(object):
     def __init__(self, hv_attribute, vm_attribute, multiplier=1, reserved=0):
         self.hv_attribute = hv_attribute
         self.vm_attribute = vm_attribute
-        self.multiplier = multiplier # TODO: Use identical units in Serveradmin.
+        self.multiplier = multiplier  # TODO Use identical units in Serveradmin
         self.reserved = reserved
 
     def __repr__(self):
@@ -151,12 +151,20 @@ class HypervisorCpuUsageLimit():
             return False
 
         hv_model = hv.dataset_obj[self.hardware_model]
+
+        # Bail out if hardware_model is not in HYPERVISOR_CPU_THRESHOLDS list
+        if hv_model not in self.hv_cpu_thresholds:
+            log.warning(
+                'Missing setting for "{}" in HYPERVISOR_CPU_THRESHOLDS'.format(
+                    hv_model))
+            return False
+
         hv_cpu_threshold = float(self.hv_cpu_thresholds[hv_model])
         hv_cpu_util_overall = hv.hv_cpu_util_overall(vm)
 
         return (
-                hv_cpu_util_overall is not None
-                and hv_cpu_util_overall > hv_cpu_threshold
+            hv_cpu_util_overall is not None and
+            hv_cpu_util_overall > hv_cpu_threshold
         )
 
 
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -9,7 +9,7 @@ from os import environ
 from tempfile import NamedTemporaryFile
 from unittest import TestCase
 
-from adminapi.dataset import Query
+from adminapi.dataset import Query, Any
 from fabric.api import env
 from fabric.network import disconnect_all
 
@@ -39,6 +39,8 @@ from igvm.puppet import clean_cert
 from igvm.settings import (
     COMMON_FABRIC_SETTINGS,
     HYPERVISOR_ATTRIBUTES,
+    HYPERVISOR_CPU_THRESHOLDS,
+    KVM_HWMODEL_TO_CPUMODEL,
     VG_NAME,
 )
 from igvm.utils import parse_size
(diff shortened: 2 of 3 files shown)


tests/test_integration.py
```diff
@@ -160,6 +161,23 @@ def check_vm_absent(self, hv_name=None):
...
+class HypervisorCpuThresholdsTest(IGVMTest):
+    """Test that all hardware_models of all hypervisors have a cpu threshold
+    value defined in HYPERVISOR_CPU_THRESHOLDS dict"""
+
+    def test_hypervisor_cpu_thresholds(self):
+        hardware_models = set([
+            x['hardware_model'] for x in Query({
+                'servertype': 'hypervisor',
+                'project': 'ndco',
+                'environment': 'production',
```
TuxPowered42, line 173: All HVs should be checked, testing ones too.

tests/test_integration.py
```diff
@@ -160,6 +161,23 @@ def check_vm_absent(self, hv_name=None):
                     )
 
 
+class HypervisorCpuThresholdsTest(IGVMTest):
+    """Test that all hardware_models of all hypervisors have a cpu threshold
+    value defined in HYPERVISOR_CPU_THRESHOLDS dict"""
+
+    def test_hypervisor_cpu_thresholds(self):
```
TuxPowered42, line 168: When we purchase new hardware we don't run igvm integration tests. So this does not necessarily have to be tested before we go to production.

However if we insist on suck check then it should be extended to cover all config variables where hardware or cpu model is checked, like KVM_HWMODEL_TO_CPUMODE.

tests/test_integration.py
```diff
@@ -160,6 +161,23 @@ def check_vm_absent(self, hv_name=None):
                     )
 
 
+class HypervisorCpuThresholdsTest(IGVMTest):
```
TuxPowered42, line 164: Inheritance from IGVMTest is not really necessary, as we're not building any VMs in this test, we only check config consistency.

meikomeik: Good point. I removed the IGVMTesting

tests/test_integration.py
meikomeik: I extended it to production and testing

tests/test_integration.py
meikomeik: I extended the class to now have two tests. One for HYPERVISOR_CPU_THRESHOLDS and one for KVM_HWMODEL_TO_CPUMODEL

I also added a setUp function to share the list of hardware_models between the two tests.
Is it correct to do it like this?

tests/test_integration.py
meikomeik: Is it correct to pass TestCase instead of IGVMTest? 
I think it is but would like a confirmation

Woellchen: `TestCase` is correct

tests/test_integration.py
Woellchen: just remove the `environment` filter altogether

igvm/hypervisor_preferences.py
```diff
@@ -151,6 +151,11 @@ def __call__(self, vm, hv) -> bool:
             return False
 
         hv_model = hv.dataset_obj[self.hardware_model]
+
+        # Bail out if hardware_model is not in HYPERVISOR_CPU_THRESHOLDS list
+        if hv_model not in self.hv_cpu_thresholds:
+            return False
```
Woellchen, line 157: will this pop up as a warning in igvm during hypervisor finding? if not i don't like silently ignoring a hv. it should be visible somehow that its not supported at all or? didnt @lacunoc have an opinion on this too?

tests/test_integration.py
Woellchen: `setUp` is good to share code between tests, yes. just keep in mind that it will still be executed once for each test. but that doesn't matter i would say

Woellchen: also good that we have integration tests for this now 👍

igvm/hypervisor_preferences.py
meikomeik: I agree that a warning would be nice here but looking at the surrounding code I see no message printed in any of the other hypervisor preferences function. 

I will test how the output looks like and report back

meikomeik: > WARNING: igvm.hypervisor_preferences: Missing setting for "Dell_R640" in HYPERVISOR_CPU_THRESHOLDS
> WARNING: igvm.hypervisor_preferences: Missing setting for "Dell_R640" in HYPERVISOR_CPU_THRESHOLDS
> INFO: igvm.hypervisor_preferences: Hypervisor "xxxxxxxxxx.xxxx.ig.local" selected using 10 preferences.
> INFO: igvm.hypervisor_preferences: Hypervisor "xxxxxxxxxx.xxxx.ig.local" selected using 10 preferences.

tests/test_integration.py
meikomeik: Resolving this comment as it seems fixed now

igvm/hypervisor_preferences.py
meikomeik: Fixed by adding a status message when a hypervisor is skipped due to missing hardware_model in HYPERVISOR_CPU_THRESHOLDS

tests/test_integration.py
```diff
@@ -160,6 +162,34 @@ def check_vm_absent(self, hv_name=None):
...
+class SettingsHardwareModelTest(TestCase):
+    """Test that all hardware_models of all hypervisors have a cpu threshold
+    value defined in HYPERVISOR_CPU_THRESHOLDS dict"""
+
+    def setUp(self):
+        self.hardware_models = set([
+            x['hardware_model'] for x in Query({
+                'servertype': 'hypervisor',
+                'project': 'ndco',
+                'environment': Any('production', 'testing'),
```
TuxPowered42, line 174: Please don't check the environment at all.

meikomeik: Sure. Deleting stuff is easy.

PR #152, opened 2020-05-27
feat(migration): Add migration log for CPU usage calculation

konstantinbest:
Since our CPU usage values of our Hypervisors are only updated once
every three hours, we might migrate multiple VMs to the same HV.

To solve this, we can integrate an igvm_migration_log to Serveradmin which will 
get updated with every migration. We will delete the log entry after
24 hours when the graphite data in Serveradmin is correct again.
We will take this migration log into account during selecting the best HV.

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -432,6 +432,23 @@ def vm_migrate(vm_hostname=None, vm_object=None, hypervisor_hostname=None,
                 _vm.start(transaction=transaction)
             _vm.reset_state()
 
+            # Add migration log entries to hypervisor and previous_hypervisor
+            hypervisor.hv_add_migration_log(_vm, '+')
+            transaction.on_rollback(
+                'reset hypervisor log',
+                hypervisor.hv_add_migration_log,
+                _vm,
+                '-',
+            )
+
+            previous_hypervisor.hv_add_migration_log(_vm, '-')
+            transaction.on_rollback(
+                'reset previous hypervisor log',
+                previous_hypervisor.hv_add_migration_log,
+                _vm,
+                '+',
+            )
+
             # Update Serveradmin
             _vm.dataset_obj['hypervisor'] = hypervisor.dataset_obj['hostname']
             _vm.dataset_obj.commit()
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -6,7 +6,7 @@ Copyright (c) 2018 InnoGames GmbH
 import logging
 import math
 from contextlib import contextmanager
-from time import sleep
+from time import sleep, time
 from xml.etree import ElementTree
 
 from igvm.vm import VM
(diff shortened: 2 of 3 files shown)


igvm/hypervisor.py
```diff
@@ -869,6 +869,43 @@ def hv_cpu_util_overall(self, vm: VM) -> float:
...
+
+        :param: none
+        :return: added cpu usage from migrations as integer
+        """
+
+        hv_migration_log = self.dataset_obj['igvm_migration_log']
+        added_cpu_usage = 0
+        for entry in hv_migration_log:
+            cpu_usage_log = entry.split(' ')[1]
+            added_cpu_usage = int(added_cpu_usage) + int(cpu_usage_log)
```
kofrezo, line 891: No need for a change here but `int(added_cpu_usage)` is not necessary because added_cpu_sage was initialised as integer and only gets summed up with integers so no need for an explicit cast here. For `cpu_usage_log` it must remain.

igvm/commands.py
```diff
@@ -432,6 +432,10 @@ def _reset_hypervisor():
                 _vm.start(transaction=transaction)
             _vm.reset_state()
 
+            # Add migration log entries to source and destination hypervisor
```
lacunoc, line 435: No need for a change, but maybe change the comment to swap source and destination or even better swap previous_hypervisor and hypervisor - which would sound more logical to me.

igvm/hypervisor.py
```diff
@@ -869,6 +869,43 @@ def hv_cpu_util_overall(self, vm: VM) -> float:
 
         hv_cpu_util_dest = self.hv_predict_vm_cpu_util(vm)
         hv_cpu_util_pct_dest = self.dataset_obj['cpu_util_pct']
-        hv_cpu_util_overall = hv_cpu_util_pct_dest + hv_cpu_util_dest
+        hv_cpu_util_log = self.hv_evaluate_migration_log()
+        hv_cpu_util_overall = hv_cpu_util_pct_dest + \
```
lacunoc, line 873: Please avoid using `\` for newlines.

igvm/hypervisor.py
```diff
@@ -869,6 +869,43 @@ def hv_cpu_util_overall(self, vm: VM) -> float:
...
+
+    def hv_add_migration_log(self, vm: VM, operator: str) -> None:
+        """HV add migration log
+
+        add the cpu usage value of a vm to the source and destination
+        hypervisors migration log
+
+        :param vm: VM object
+        :param operator: plus or minus for source and dest HV
+        :return: none
```
lacunoc, line 903: Just remove it from the docstring, when there's no return value.

igvm/hypervisor.py
```diff
@@ -869,6 +869,43 @@ def hv_cpu_util_overall(self, vm: VM) -> float:
...
+                              hv_cpu_util_dest + hv_cpu_util_log
 
         return float(hv_cpu_util_overall)
+
+    def hv_evaluate_migration_log(self) -> int:
+        """HV evaluate migration log
+
+        Calculate the added cpu usage from previous migrations
+
+        :param: none
```
lacunoc, line 883: Just remove it from the docstring, when there aren't any params.

igvm/commands.py
```diff
@@ -432,6 +432,10 @@ def _reset_hypervisor():
                 _vm.start(transaction=transaction)
             _vm.reset_state()
 
+            # Add migration log entries to hypervisor and previous_hypervisor
+            hypervisor.hv_add_migration_log(_vm, '+')
```
Woellchen, line 436: we are missing transaction rollback callbacks here. if for whatever reason something is failing or the vm commit below is failing, we will end up with invalid migration log entries on serveradmin

Woellchen:
> Only concern from my side is design wise. Maybe the evaluation should be implemented as hypervisor preference rather than as methods on the hypervisor itself. What do you think @TuxPowered42 about this ?
the cpu utilization is already implemented as a preference. he just extended the calculation

igvm/commands.py
konstantinbest: Added the rollback callbacks.

igvm/hypervisor.py
```diff
@@ -869,6 +869,44 @@ def hv_cpu_util_overall(self, vm: VM) -> float:
...
+        add the cpu usage value of a vm to the source and destination
+        hypervisors migration log
+
+        :param vm: VM object
+        :param operator: plus or minus for source and dest HV
+        """
+
+        cpu_usage_vm = self.hv_predict_vm_cpu_util(vm)
+        timestamp = int(time())
+        log_entry = '{} {}{}'.format(timestamp, operator, round(cpu_usage_vm))
```
TuxPowered42, line 909: Is a multiattribute with spaces working fine in Serveradmin? I assume it is through API, but will it show fine in webinterface?

konstantinbest: It is also working in the webinterface. You can check SA testing, there are some HV which have this attribute already set for the tests.

konstantinbest: It is working in the webinterface.

konstantinbest:
retest this please

konstantinbest:
retest this please

konstantinbest:
retest this please

konstantinbest:
We have problems with the test, but verified it manually.

PR #153, opened 2020-06-02
feat(migration): Add unittest for igvm_migration_log

lacunoc:


lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
Unfortunately, the testing framework is totally unreliable atm.
The test passed, when I only tested the new unittest explicitly https://sysadmin-jenkins.innogames.de/job/integration/job/igvm-testing/105/console

tests/test_integration.py
```diff
@@ -455,6 +456,21 @@ def test_info(self):
         self.vm.shutdown()
         host_info(VM_HOSTNAME)
 
+    @patch('igvm.vm.VM.vm_performance_value', return_value=5.0)
+    def test_igvm_migration_log(self, mock_vm_performance_value):
+        regex = r'\A([0-9]{10}\s[-|+][0-9]{1,3})+\Z'
+
+        for hv in self.hvs:
+            if hv.dataset_obj['hostname'] == self.vm.dataset_obj['hypervisor']:
+                hv.dataset_obj['igvm_migration_log'] = ''
```
kofrezo, line 465: igvm_migration_log is a multi attribute which implements the default python set interface. While it is supported to set it to an empty string to remove all values it is nicer to use the clear() method to make the reader aware that this a multi attribute.

tests/test_integration.py
```diff
@@ -455,6 +456,21 @@ def test_info(self):
...
+                hv.dataset_obj.commit()
+                self.assertEqual(str(hv.dataset_obj['igvm_migration_log']), '')
```
kofrezo, line 467: Evaluating the boolean representation directly instead of the string representation is possible.
```
>>> h['igvm_migration_log']
MultiAttr()
>>> bool(h['igvm_migration_log'])
False
>>> h['igvm_migration_log'].add('1591169717 +5')
>>> bool(h['igvm_migration_log'])
True
```

tests/test_integration.py
```diff
@@ -455,6 +456,21 @@ def test_info(self):
...
+
+                hv.hv_add_migration_log(self.vm, '-')
+                hv.hv_add_migration_log(self.vm, '+')
+                for entry in hv.dataset_obj['igvm_migration_log']:
+                    self.assertRegex(entry, regex)
```
kofrezo, line 472: This should be already enforced by Serveradmin but it does not harm I guess.

Woellchen:
test this please

Woellchen:
test this please

tests/test_integration.py
Woellchen: i think we should even check for the exact expected migration log entry, not for any regex

Woellchen:
test this please

tests/test_integration.py
kofrezo: Good point.

tests/test_integration.py
lacunoc: Using clear() now.

tests/test_integration.py
lacunoc: Thats right. I'm testing the bool now.

tests/test_integration.py
lacunoc: I reworked it, checking the log on the source hypervisor, migrating the VM and checking the log on the destination hypervisor then.

tests/test_integration.py
```diff
@@ -455,6 +457,36 @@ def test_info(self):
...
+        self.assertFalse(
+            bool(self.vm.hypervisor.dataset_obj['igvm_migration_log'])
+        )
+
+        cpu_usage_vm = self.vm.hypervisor.hv_predict_vm_cpu_util(self.vm)
+        timestamp = int(time())
+        self.vm.hypervisor.hv_add_migration_log(self.vm, '-')
+        self.assertEqual(
+            list(self.vm.hypervisor.dataset_obj['igvm_migration_log']),
+            ['{} -{}'.format(timestamp, round(cpu_usage_vm))])
```
konstantinbest, line 473: If hv_add_migration_log takes more than one second, the timestamp will be different.

tests/test_integration.py
```diff
@@ -455,6 +457,36 @@ def test_info(self):
...
+    def test_igvm_migration_log(self, mock_vm_performance_value):
+        self.vm.hypervisor.dataset_obj['igvm_migration_log'].clear()
+        self.vm.hypervisor.dataset_obj.commit()
+        self.assertFalse(
+            bool(self.vm.hypervisor.dataset_obj['igvm_migration_log'])
+        )
+
+        cpu_usage_vm = self.vm.hypervisor.hv_predict_vm_cpu_util(self.vm)
+        timestamp = int(time())
+        self.vm.hypervisor.hv_add_migration_log(self.vm, '-')
```
Woellchen, line 470: unfortunately this does not make any sense now. we want to test that `vm_migrate()` add this entry. this test will always pass now. don't write the migration log in the test. do the migration and then do the `assertEqual`.
same applies for the other `hv_add_migration_log()` below.
save `self.vm.hypervisor` into some `prev_hv` variable because the `vm_migrate()` call will overwrite it. then you can do your asserts on `prev_hv` and `self.vm.hypervisor`

Woellchen: oh and to circumvent the problem with the timestamp you can mock the current time as well

tests/test_integration.py
lacunoc: timestamp is now mocked

tests/test_integration.py
lacunoc: I reworked it again.
For whatever reason the VM object isn't reinitialized after the migration, so I had to workaround it a bit by getting a fresh object with `_get_vm(VM_HOSTNAME)`

Anyway, only the migration is now manipulating the `igvm_migration_log` and the asserts are working as expected.

PR #154, opened 2020-06-23
feat(build/aws): Spawn VM with a fitting vm_type

lacunoc:
We can't ensure, that the commited aws_instance_type in SA
is still available in the corresponding region of AWS.
To avoid breaking builds we use our new vm_performance_factor
to get a list of fitting vm_types in AWS, which have at least -25%
or at max +25% the performance of the original VM.
We try the next fitting one, if the one before is not available
anymore.

### DIFF
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -5,25 +5,34 @@ Copyright (c) 2018 InnoGames GmbH
 import json
 import logging
 import os
-import random
 import time
 from base64 import b64decode
 from hashlib import sha1, sha256
 from io import BytesIO
+from pathlib import Path
 from re import compile as re_compile
-from typing import Optional
+from typing import Optional, List, Union
 from uuid import uuid4
 
 import boto3
 import tqdm
-from botocore.exceptions import ClientError
+from botocore.exceptions import ClientError, CapacityNotAvailableError
 from fabric.api import cd, get, hide, put, run, settings
 from fabric.contrib.files import upload_template
 from fabric.exceptions import NetworkError
+from json.decoder import JSONDecodeError
+from urllib.request import Request, urlopen
+from urllib.error import HTTPError
 
 from igvm.exceptions import ConfigError, RemoteCommandError, VMError
 from igvm.host import Host
-from igvm.settings import AWS_RETURN_CODES, COMMON_FABRIC_SETTINGS
+from igvm.settings import (
+    AWS_ECU_FACTOR,
+    AWS_RETURN_CODES,
+    AWS_INSTANCES_OVERVIEW_FILE,
+    AWS_INSTANCES_OVERVIEW_FILE_etag,
+    AWS_INSTANCES_OVERVIEW_URL,
+)
 from igvm.transaction import Transaction
 from igvm.utils import parse_size, wait_until
 from igvm.puppet import clean_cert
@@ -502,36 +511,46 @@ class VM(Host):
 
         ec2 = boto3.client('ec2')
 
-        try:
-            response = ec2.run_instances(
-                ImageId=self.dataset_obj['aws_image_id'],
-                InstanceType=self.dataset_obj['aws_instance_type'],
-                KeyName=self.dataset_obj['aws_key_name'],
-                SecurityGroupIds=list(
-                    self.dataset_obj['aws_security_group_ids']),
-                SubnetId=self.dataset_obj['aws_subnet_id'],
-                Placement={
-                    'AvailabilityZone': self.dataset_obj['aws_placement']
-                },
-                PrivateIpAddress=str(self.dataset_obj['intern_ip']),
-                UserData='' if postboot is None else postboot,
-                TagSpecifications=[
-                    {
-                        'ResourceType': 'instance',
-                        'Tags': [
-                            {
-                                'Key': 'Name',
-                                'Value': self.dataset_obj['hostname'],
-                            },
-                        ]
+        vm_types_overview = self.aws_get_instances_overview()
+        if vm_types_overview:
+            vm_types = self.aws_get_fitting_vm_types(vm_types_overview)
+        else:
+            vm_types = [self.dataset_obj['aws_instance_type']]
+
+        for vm_type in vm_types:
+            try:
+                response = ec2.run_instances(
+                    ImageId=self.dataset_obj['aws_image_id'],
+                    InstanceType=vm_type,
+                    KeyName=self.dataset_obj['aws_key_name'],
+                    SecurityGroupIds=list(
+                        self.dataset_obj['aws_security_group_ids']),
+                    SubnetId=self.dataset_obj['aws_subnet_id'],
+                    Placement={
+                        'AvailabilityZone': self.dataset_obj['aws_placement']
                     },
-                ],
-                DryRun=False,
-                MinCount=1,
-                MaxCount=1)
-            log.info(response)
-        except ClientError as e:
-            raise VMError(e)
+                    PrivateIpAddress=str(self.dataset_obj['intern_ip']),
+                    UserData='' if postboot is None else postboot,
+                    TagSpecifications=[
+                        {
+                            'ResourceType': 'instance',
+                            'Tags': [
+                                {
+                                    'Key': 'Name',
+                                    'Value': self.dataset_obj['hostname'],
+                                },
+                            ]
+                        },
+                    ],
+                    DryRun=False,
+                    MinCount=1,
+                    MaxCount=1)
+                log.info(response)
+                break
+            except ClientError as e:
+                raise VMError(e)
+            except CapacityNotAvailableError as e:
+                continue
 
         if run_puppet:
             self.run_puppet(clear_cert=True, debug=debug_puppet)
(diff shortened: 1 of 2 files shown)


lacunoc:
retest this please

lacunoc:
retest this please

igvm/vm.py
```diff
@@ -849,3 +866,83 @@ def vm_performance_value(self) -> float:
...
+            overview = json.load(f)
+            f.close()
+        else:
+            r = requests.get(url, timeout=timeout)
+            if r.status_code == requests.codes.ok:
+                overview = r.json()
+            else:
+                overview = None
+
+        return overview
```
kofrezo, line 897: You can return early:

```
...
888 return json.load(f)
...
893 return r.json()
...
895 return None
```

igvm/vm.py
```diff
@@ -849,3 +866,83 @@ def vm_performance_value(self) -> float:
...
+        vm_types for this targeted ECU range.
+
+        :param: overview: instances.json with a complete
+                          instance_types overview
+
+        :return: Fitting VM types as list
+        """
+
+        vm_performance_value = self.vm_performance_value()
+        region = self.dataset_obj['aws_placement'][:-1]
```
kofrezo, line 914: If that is a possible case: What happens when aws_placment attribute is empty (IndexError) ?

lacunoc: It's not possible. The placement gets inherited by the network (project_network), which will be/is already prepared in advance. So if you're moving a VM network wise or create a new one within this network, you'll automatically get the needed placement.

igvm/vm.py
```diff
@@ -5,25 +5,32 @@
 import json
 import logging
 import os
-import random
+import requests
```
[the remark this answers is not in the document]
lacunoc: Changed to `urllib.request.urlopen`

igvm/vm.py
```diff
@@ -849,3 +866,83 @@ def vm_performance_value(self) -> float:
...
+                 or None, if load/download failed
+        """
+
+        url = AWS_INSTANCES_OVERVIEW_URL
+        file = Path(AWS_INSTANCES_OVERVIEW_FILE)
+
+        if file.exists():
+            f = open(AWS_INSTANCES_OVERVIEW_FILE)
+            overview = json.load(f)
+            f.close()
```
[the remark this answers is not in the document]
lacunoc: Using context manager now

igvm/vm.py
lacunoc: True, changed!

Woellchen:
retest this please

igvm/vm.py
```diff
@@ -849,3 +866,81 @@ def vm_performance_value(self) -> float:
...
+        :param: timeout: Timeout value for the get request
+
+        :return: VM types overview as list
+                 or None, if load/download failed
+        """
+
+        url = AWS_INSTANCES_OVERVIEW_URL
+        file = Path(AWS_INSTANCES_OVERVIEW_FILE)
+
+        if file.exists():
```
Woellchen, line 887: what if the content of the file changes? we will never try to re-download it right?

igvm/vm.py
```diff
@@ -849,3 +866,81 @@ def vm_performance_value(self) -> float:
...
+        ecu_targeted = {
+            'min':
+                (vm_performance_value * AWS_ECU_FACTOR) -
+                (vm_performance_value * AWS_ECU_FACTOR * 0.25),
+            'max':
+                (vm_performance_value * AWS_ECU_FACTOR) +
+                (vm_performance_value * AWS_ECU_FACTOR * 0.25)
+        }
+
+        nested_dict = (
```
Woellchen, line 923: what exactly do you want to achieve with this construct? i dont quite get it

igvm/vm.py
lacunoc: Additionally to this PR there will be a new cronjob on our controls, which is doing a daily download of this file

igvm/vm.py
lacunoc: Using a `defaultdict()` is not sufficient here:
```
>>> vm_types = defaultdict()
>>> vm_types['foo']['foo2'] = 'bar'
  File "<stdin>", line 1, in <module>
KeyError: 'foo'
>>> 
```

I want to be able to create a nested dictionary, with the instance_type name as the key with multiple sub-keys and their corresponding values
```
>>> nested_dict = lambda: defaultdict(nested_dict)
>>> vm_types = nested_dict()
>>> vm_types['foo']['foo2'] = 'bar'
>>> vm_types
defaultdict(<function <lambda> at 0x7f064c426310>, {'foo': defaultdict(<function <lambda> at 0x7f064c426310>, {'foo2': 'bar'})})
>>> vm_types['foo']['foo2']
'bar'
```

It would look like this:
`defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'m5.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'm5.2xlarge', 'memory': 32.0, 'ECU': 37.0, 'price': 0.46}), 'i2.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'i2.2xlarge', 'memory': 61.0, 'ECU': 27.0, 'price': 2.026}), 'i3en.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'i3en.2xlarge', 'memory': 64.0, 'ECU': 37.0, 'price': 1.08}), 'p3.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'p3.2xlarge', 'memory': 61.0, 'ECU': 31.0, 'price': 3.823}), 'r5.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'r5.2xlarge', 'memory': 64.0, 'ECU': 37.0, 'price': 0.608}), 'm5d.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'm5d.2xlarge', 'memory': 32.0, 'ECU': 37.0, 'price': 0.544}), 'c5n.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'c5n.2xlarge', 'memory': 21.0, 'ECU': 39.0, 'price': 0.492}), 'c5.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'c5.2xlarge', 'memory': 16.0, 'ECU': 39.0, 'price': 0.388}), 'c5d.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'c5d.2xlarge', 'memory': 16.0, 'ECU': 39.0, 'price': 0.444}), 'c3.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'c3.2xlarge', 'memory': 15.0, 'ECU': 28.0, 'price': 0.516}), 'r4.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'r4.2xlarge', 'memory': 61.0, 'ECU': 31.0, 'price': 0.6402}), 'd2.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'd2.2xlarge', 'memory': 61.0, 'ECU': 28.0, 'price': 1.588}), 'c4.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'c4.2xlarge', 'memory': 15.0, 'ECU': 31.0, 'price': 0.454}), 'r5d.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'r5d.2xlarge', 'memory': 64.0, 'ECU': 37.0, 'price': 0.692}), 'i3.2xlarge': defaultdict(<function VM.aws_get_fitting_vm_types.<locals>.<lambda> at 0x7fc7d04130d0>, {'instance_type': 'i3.2xlarge', 'memory': 61.0, 'ECU': 31.0, 'price': 0.744})})`

igvm/vm.py
Woellchen: i would not add more complexity to this. we really don't need a cronjob. instead we can make use of the http protocol and it's cache headers. just try a `curl -I https://www.ec2instances.info/instances.json`. you will see an `etag` header which is kind of a fingerprint. so i would do it the following way:
- if file does not exist, make a GET request. save the etag from the headers into a separate file, maybe AWS_INSTANCES_OVERVIEW_FILE.etag.
- if file does exist, make a HEAD request and compare the etags. if they differ, repeat the first step. in case the HEAD request fails you can decide whether you want to abort or just ignore and use the existing file in any case (i would still log it so we have a chance to detect problems)

igvm/vm.py
Woellchen: ah i see. i think this is a very complicated alias to just doing this:
```
vm_types = dict()

for t in overview:
    vm_types[t['instance_type']] = dict()
    vm_types[t['instance_type']]['instance_type'] = t['instance_type']
    vm_types[t['instance_type']]['memory'] = t['memory']
   ...
```
at least i would find that way more readable and clear

igvm/vm.py
TuxPowered42: > Additionally to this PR there will be a new cronjob on our controls, which is doing a daily download of this file

Igvm can't be dependent on control servers or cron jobs. It's a tool on it's own and must be working fine from peoples' laptops. Please download files when it is run, just like we do with images.

lacunoc: The download, if the file was not there, was already in, but I reworked the function, like @Woellchen suggested.

igvm/vm.py
lacunoc: Absolutely right, got rid of it. Thanks!

igvm/vm.py
```diff
@@ -849,3 +866,94 @@ def vm_performance_value(self) -> float:
...
+        url = AWS_INSTANCES_OVERVIEW_URL
+        file = Path(AWS_INSTANCES_OVERVIEW_FILE)
+        etag_file = Path(AWS_INSTANCES_OVERVIEW_FILE_etag)
+
+        head_req = Request(url, method='HEAD')
+        resp = urlopen(head_req, timeout=timeout)
+        if resp.status == 200:
+            etag = dict(resp.info())['ETag']
+        else:
+            log.debug('Could not retrieve ETag from {}'.format(url))
```
Woellchen, line 891: this should be a warning. otherwise nobody sees it

igvm/vm.py
```diff
@@ -849,3 +866,94 @@ def vm_performance_value(self) -> float:
...
+                    prev_etag = f.read()
+                if etag and etag == prev_etag:
+                    with open(AWS_INSTANCES_OVERVIEW_FILE) as f:
+                        return json.load(f)
+
+        resp = urlopen(url, timeout=timeout)
+        if etag:
+            with open(AWS_INSTANCES_OVERVIEW_FILE_etag, 'w') as f:
+                f.write(etag)
+        with open(AWS_INSTANCES_OVERVIEW_FILE, 'w') as f:
```
Woellchen, line 905: lets combine those two `open` calls.
```
content = resp.read().decode('utf-8')
f.write(content)
return json.loads(content)
```

igvm/vm.py
```diff
@@ -849,3 +866,94 @@ def vm_performance_value(self) -> float:
...
+
+        :return: VM types overview as list
+        """
+
+        url = AWS_INSTANCES_OVERVIEW_URL
+        file = Path(AWS_INSTANCES_OVERVIEW_FILE)
+        etag_file = Path(AWS_INSTANCES_OVERVIEW_FILE_etag)
+
+        head_req = Request(url, method='HEAD')
+        resp = urlopen(head_req, timeout=timeout)
```
Woellchen, line 887: if you want this graceful i would wrap it into a proper try..except block. otherwise it can still abort the vm build

igvm/vm.py
```diff
@@ -849,3 +866,94 @@ def vm_performance_value(self) -> float:
...
+        if resp.status == 200:
+            etag = dict(resp.info())['ETag']
+        else:
+            log.debug('Could not retrieve ETag from {}'.format(url))
+
+        if file.exists():
```
Woellchen, line 893: shortcut: `if file.exists() and etag_file.exists()`. i would also add a `and etag`. this will prevent unnecessary file loading. you can shorten the `if etag and etag == prev_etag:` then accordingly

igvm/vm.py
```diff
@@ -849,3 +866,94 @@ def vm_performance_value(self) -> float:
...
+            if t['memory'] >= (self.dataset_obj['memory'] / 1024):
+                vm_types[t['instance_type']]['memory'] = t['memory']
+            if 'ECU' in t and isinstance(t['ECU'], (int, float)):
+                if ecu_targeted['min'] <= t['ECU'] <= ecu_targeted['max']:
+                    vm_types[t['instance_type']]['ECU'] = t['ECU']
+            if region in t['pricing']:
+                vm_types[t['instance_type']]['price'] = float(
+                    t['pricing'][region]['linux']['ondemand'])
+
+            if len(vm_types[t['instance_type']]) < 4:
```
Woellchen, line 951: sorry i did not catch this earlier. i would rewrite this loop a little bit because it is kinda unclear at first what this condition is really for. we can also prevent unnecessarily adding and again removing entries from the main dict.
```
for t in overview:
    if region not in t['pricing']:
        continue
    if t['memory'] < (self.dataset_obj['memory'] / 1024):
        continue
    if 'ECU' not in t or not isinstance(t['ECU'], (int, float)):
        continue

    # now it gets very straightforward:
    vm_types[t['instance_type']] = dict(
        instance_type=t['instance_type'],
        memory=t['memory'],
        ECU=t['ECU'],
        price=float(t['pricing'][region]['linux']['ondemand']),
    )
```

igvm/vm.py
Woellchen: cool, the functionality is now good. except for small remarks. see my other comments for that

igvm/vm.py
Woellchen: actually the last statement could be as simple as `vm_types[t['instance_type']] = t` if you like

igvm/vm.py
lacunoc: Changed to warning

igvm/vm.py
lacunoc: Changed

igvm/vm.py
lacunoc: I wrapped it and hopefully caught all possible Exceptions here.

igvm/vm.py
lacunoc: Changed accordingly

igvm/vm.py
lacunoc: That's actual even nicer. Thanks for the cleanup:)

PR #157, opened 2020-08-20
feat(build/migrate): Add HV env preference and enforce-env parameter

lacunoc:
We want to prefer hypervisors of the same environment as build/migration destination. Since not all projects have testing hypervisors, this is just a preference, which will not affect the hypervisor selection. Additionally there is also a parameter now, to be able to enforce it for builds/migrations.

### DIFF
--- a/igvm/cli.py
+++ b/igvm/cli.py
@@ -177,6 +177,12 @@ def parse_args():
             ' operator to shut down VM.'
         ),
     )
+    subparser.add_argument(
+        '--enforce-vm-env',
+        dest='enforce_vm_env',
+        action='store_true',
+        help='Build or migrate VM only to a HV with the same environment of VM'
+    )
 
     subparser = subparsers.add_parser(
         'change-address',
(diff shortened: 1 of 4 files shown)


igvm/cli.py
```diff
@@ -177,6 +177,12 @@ def parse_args():
             ' operator to shut down VM.'
         ),
     )
+    subparser.add_argument(
+        '--enforce-env',
+        dest='enforce_env',
+        action='store_true',
+        help='Build or migrate VM only to a HV with the same environment'
```
kofrezo, line 184: Maybe write same environment of VM to avoid miss understandings.

igvm/cli.py
```diff
@@ -177,6 +177,12 @@ def parse_args():
             ' operator to shut down VM.'
         ),
     )
+    subparser.add_argument(
+        '--enforce-env',
```
kofrezo, line 181: Maybe rename to `enforce-vm-env` ?

kofrezo (approved):

kofrezo:
I put some changes on top which keep the new behaviour of environment selection but always use the variable from IGVM_MODE because this is a thing we need for functional and integration tests.

igvm/cli.py
lacunoc: added

igvm/cli.py
lacunoc: renamed

kofrezo:
test this please

PR #161, opened 2020-08-24
Best hypervisor timeout

kofrezo:


kofrezo:
This is no longer required because the improvments made to the hypervisor selection reduced the selection speed to almost instant.

PR #162, opened 2020-08-24
Test user

kofrezo:


kofrezo:
This has been reworked and recently adjusted for a different feature with https://github.com/innogames/igvm/pull/157

PR #165, opened 2020-08-25
Revert "Rework the VM renaming function"

kofrezo:
Reverts innogames/igvm#142


lacunoc (approved):

kofrezo:
This can never succeed in the test pipeline as it was broken from the beginning on (parallelisation issue in the tests)

PR #166, opened 2020-08-25
Restore fix for vm_rename function

kofrezo:
The added test for the vm_rename fix in https://github.com/innogames/igvm/pull/142 introduced a bug where the cleanup task failed to delete the renamed VMs when starting the tests which only became visible in follow up tests.

This PR brings back the origin fix for the vm_rename and a work around to allow the cleanup task to delete the VMs.


kofrezo:
This is not optimal as it still can happen that we fail at the vm_rename test and have to delete the VM manually but I think this is good enough for now. We can aim for adjusting the pattern to delete VMs in a follow up PR.

lamaral (approved):
This makes sense. 

I think the chances of failing on vm_rename are somewhat slim, since we very rarely touch it, so this can get us going for a while.

PR #171, opened 2020-09-22
Rework HV selection algorithm

Woellchen:
This is a rework of the hv selection algorithm. It changes the way the preferences work.
Preferences from now on do not return arbitrary numbers or other types but "scores" in the range of 0.0 - 1.0. Booleans are allowed for convenience, but translate to either 0.0 or 1.0, respectively. A score of 0.0 means that the hv will be excluded.
All preferences are evaluated and a total score is calculated, which is also in the range of 0.0 - 1.0.
HVs are then sorted based on the highest score, from high to low. The highest score represents the best fitting hv for a given VM.
This PR also fixes a few issues with the existing preferences.
From my tests I can say that the distribution should definitely improve with this, also it is less likely that similar VMs end up on the same HV.
This is still a work in progress but the general idea and workings of this are done, so I would appreciate some early feedback :)

### DIFF
--- a/igvm/hypervisor_preferences.py
+++ b/igvm/hypervisor_preferences.py
@@ -348,13 +348,17 @@ class PreferenceEvaluator:
             # We expect normalized values from 0 - 1.
             if result < 0. or result > 1.:
                 raise ValueError(
-                    'Preference must be expressed in a 0.0 - 1.0 range, '
-                    '{} given.'.format(result)
+                    'Preference "{}" for Hypervisor "{}" must be expressed '
+                    'in a 0.0 - 1.0 range, {} given.'.format(
+                        str(pref),
+                        str(hv),
+                        result,
+                    )
                 )
 
             # Add up the individual preference scores.
             if result > 0.:
-                log.debug('Preference {} matches with score {:.4f}.'.format(
+                log.debug('Preference "{}" matches with score {:.4f}.'.format(
                     str(pref),
                     result,
                 ))
--- a/igvm/hypervisor_preferences.py
+++ b/igvm/hypervisor_preferences.py
@@ -59,6 +59,8 @@ class InsufficientResource(HypervisorPreference):
     def get_score(self, vm, hv) -> Union[float, bool]:
         # Treat freshly created HVs always passing this check
         if not hv.dataset_obj[self.hv_attribute]:
+            # Because new installed Hypervisors don't have values 
+            # (e.g. for load) - they are fresh meat ...
             return True
 
         # Calculate the remaining "size" of the resource
--- a/igvm/hypervisor_preferences.py
+++ b/igvm/hypervisor_preferences.py
@@ -18,7 +18,7 @@ log = getLogger(__name__)
 class HypervisorPreference(object):
     """The base class for all HV preferences."""
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         """Calculates a preference value to indicate how good a HV fits.
 
         @param vm: The VM object to check against the HV.
@@ -55,7 +55,7 @@ class InsufficientResource(HypervisorPreference):
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         # Treat freshly created HVs always passing this check
         if not hv.dataset_obj[self.hv_attribute]:
             return True
@@ -81,10 +81,11 @@ class InsufficientResource(HypervisorPreference):
 class OtherVMs(HypervisorPreference):
     """Count the other VMs on the hypervisor with the same attributes."""
 
-    def __init__(self, attributes=[], values=None) -> None:
+    def __init__(self, attributes: list, values: list = None) -> None:
         assert values is None or len(attributes) == len(values)
-        self.attributes = attributes
-        self.values = values
+
+        self.attributes: list = attributes
+        self.values: list = values
 
     def __repr__(self) -> str:
         args = ''
@@ -95,7 +96,7 @@ class OtherVMs(HypervisorPreference):
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         # If there are no VMs at all that makes a perfect match.
         if len(hv.dataset_obj['vms']) == 0:
             return 1.
@@ -155,7 +156,7 @@ class HypervisorAttributeValue(HypervisorPreference):
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         value = hv.dataset_obj[self.attribute]
 
         # If there is no value we assume it's a fresh HV.
@@ -180,7 +181,7 @@ class HypervisorAttributeValueLimit(HypervisorPreference):
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         value = hv.dataset_obj[self.attribute]
 
         # If there is no value we assume it's a fresh HV.
@@ -213,7 +214,7 @@ class HypervisorCpuUsageLimit(HypervisorPreference):
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         # New VM has no hypervisor attribute, yet, so we cannot calculate a
         # score here. We will just allow all HVs to take that VM for now.
         if not vm.hypervisor:
@@ -263,7 +264,7 @@ class HypervisorEnvironmentValue(HypervisorPreference):
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         hypervisor_env = hv.dataset_obj[self.hv_env]
         vm_env = vm.dataset_obj['environment']
 
@@ -288,7 +289,7 @@ class OverAllocation(HypervisorPreference):
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def get_preference(self, vm, hv) -> Union[float, bool]:
+    def get_score(self, vm, hv) -> Union[float, bool]:
         # Calculate the current HVs overbooking "level".
         cur_hv_cpus = sum(
             v[self.attribute] for v in vm.hypervisor.dataset_obj['vms']
@@ -316,10 +317,14 @@ class OverAllocation(HypervisorPreference):
 
 
 class PreferenceEvaluator(object):
+    """Evaluates all preferences for a given VM and HV and calculates the total
+    score based on which the most preferred HVs can be picked.
+    """
     def __init__(self, preferences: List[HypervisorPreference]) -> None:
         self.preferences = preferences
 
-    def get_preference(self, vm, hv) -> float:
+    def get_total_score(self, vm, hv) -> float:
+        """Calculates the total score for a given VM and HV pair."""
         n_prefs = len(self.preferences)
         matched_prefs = 0
         sum_prefs = 0.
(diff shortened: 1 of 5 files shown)


igvm/hypervisor_preferences.py
```diff
@@ -66,209 +96,336 @@ def __repr__(self):
...
 
+        # If the environment is matching, this is our candidate!
         if hypervisor_env == vm_env:
-            return True
+            return 1.
+
+        # Unfortunately we don't always have enough HVs to ensure the same
+        # environment. Therefore we will just highly discourage the hv
+        # instead of excluding it.
+        return 0.01
```
TuxPowered42, line 278: I believe that `environment` should be a hard criterion. Not having enough HVs is an issue that needs to be addressed by reassigning or moving hardware around, not by moving VMs.

igvm/hypervisor_preferences.py
```diff
@@ -66,209 +96,336 @@ def __repr__(self):
...
+        rel_overbooking = tgt_ovr_allc / cur_ovr_allc
+        if rel_overbooking > 1.:
+            return .01
 
-class HashDifference(object):
-    """Return some arbitrary number to have stable ordering"""
-    def __repr__(self):
-        return '{}()'.format(type(self).__name__)
+        # Normalize the value. This is no hard criteria, except for non-
+        # overbookable resources like disk. Hence we can allow overbooked cpus.
```
TuxPowered42, line 313: This is true, but the logic is opposite. There is hard criteria for majority of resources (memory and storage pool), the CPU is the exception.

igvm/hypervisor_preferences.py
```diff
@@ -66,209 +96,336 @@ def __repr__(self):
...
+
+                matched_prefs += 1
+                sum_prefs += result
+            else:
+                log.debug(
+                    'Hypervisor "{}" is skipped because preference "{}" does '
+                    'not match.'.format(str(hv), str(pref)),
+                )
+
+        # Exclude HV if only one criteria has failed.
```
TuxPowered42, line 360: If at least one? If even one? If any?

igvm/hypervisor_preferences.py
Woellchen: but in af for example we don't have a single staging hv. we have a parameter to enforce this: `--enforce-vm-env`. this should not be a hard criterion imo

igvm/hypervisor_preferences.py
Woellchen: if any of the preferences fails, then yes. hv will be excluded

igvm/hypervisor_preferences.py
TuxPowered42: I don't think people will be very happy if I reboot a HV just because it's not production and there are production VMs on it.

igvm/hypervisor_preferences.py
TuxPowered42: Then please correct the way that sentence is written.

igvm/hypervisor_preferences.py
```diff
@@ -12,30 +12,43 @@
...
 from logging import getLogger
-
-from igvm.utils import LazyCompare
+from typing import Union, List, Optional
 
 log = getLogger(__name__)
 
 
-class InsufficientResource(object):
+class HypervisorPreference(object):
```
kofrezo, line 18: Passing `object` is optional in Python3. If not specified everything implicitly inherits from object.

igvm/hypervisor_preferences.py
```diff
@@ -12,30 +12,43 @@
...
+    def __repr__(self) -> str:
         args = repr(self.hv_attribute)
         if self.reserved:
             args += ', reserved=' + repr(self.reserved)
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def __call__(self, vm, hv):
+    def get_preference(self, vm, hv) -> Union[float, bool]:
         # Treat freshly created HVs always passing this check
```
kofrezo, line 59: With False being evaluated to `0.0` it would not longer pass this check but fail no ?

igvm/hypervisor_preferences.py
```diff
@@ -66,33 +86,45 @@ def __repr__(self):
...
-    def __call__(self, vm, hv):
+    def get_preference(self, vm, hv) -> Union[float, bool]:
         result = 0
         for other_vm in hv.dataset_obj['vms']:
             if other_vm['hostname'] == vm.dataset_obj['hostname']:
                 continue
+
             if self.values and not all(
                 vm.dataset_obj[a] == v
                 for a, v in zip(self.attributes, self.values)
```
kofrezo, line 114: Should we rename this to something more speaking like `attr` too ?

igvm/hypervisor_preferences.py
```diff
@@ -47,17 +60,24 @@ def __call__(self, vm, hv):
...
         remaining_size = total_size - vms_size - self.reserved
 
-        return remaining_size < vm.dataset_obj[self.vm_attribute]
+        # does not fit at all
+        vm_size = vm.dataset_obj[self.vm_attribute]
+        if remaining_size < vm_size:
+            return False
 
+        # normalize the resource usage
+        return 1 - (vm_size / remaining_size)
```
kofrezo, line 78: If there is a slight chance that this becomes a integer we should better cast it to float.

igvm/hypervisor_preferences.py
```diff
@@ -101,41 +133,58 @@ def __repr__(self):
...
         value = hv.dataset_obj[self.attribute]
 
-        return value is not None, value
+        # if there is no value we assume it's a fresh hv
+        if value is None:
+            return 1.
 
+        # normalize the value. this is only valid for "exhaution values" like
+        # cpu_util_pct and iops_avg
+        return 1 - (value / 100)
```
kofrezo, line 169: If there is a slight chance that this becomes a integer we should better cast it to float.

igvm/hypervisor_preferences.py
```diff
@@ -101,41 +133,58 @@ def __repr__(self):
...
+        # if there is no value we assume it's a fresh hv
+        if value is None:
+            return 1.
+
+        if value > self.limit:
+            return False
 
+        # normalize the value. this is only valid for "exhaution values" like
+        # cpu_util_pct and iops_avg
+        return 1 - (value / 100)
```
kofrezo, line 198: If there is a slight chance that this becomes a integer we should better cast it to float.

igvm/hypervisor_preferences.py
```diff
@@ -162,18 +211,27 @@ def __call__(self, vm, hv) -> bool:
...
+            return 1.
+
+        # since this is a limiting preference, we don't want any vm end up on
+        # the hv that would exceed the cpu threshold
+        if hv_cpu_util_overall > hv_cpu_threshold:
+            return False
 
+        # normalize the value. this is only valid for "exhaution values" like
+        # cpu_util_pct and iops_avg
+        return 1 - (hv_cpu_util_overall / hv_cpu_threshold)
```
kofrezo, line 249: If there is a slight chance that this becomes a integer we should better cast it to float.

igvm/hypervisor_preferences.py
```diff
@@ -219,16 +278,104 @@ def __call__(self, vm, hv):
...
+    def get_preference(self) -> float:
+        if self.total:
+            return self.total
+
+        n_prefs = len(self.preferences)
+        matched_prefs = 0
+        sum_prefs = 0.
+
+        for p in self.preferences:
+            result = float(p.get_preference(self.vm, self.hv))
```
kofrezo: Methods would fail before if they do not return a float because of the type hints.

kofrezo (changes requested):

igvm/hypervisor_preferences.py
Woellchen: i would like to get this version of the pr merged. i have not changed any logic here. it was the same before. let's create a ticket for this and then we can think about how to solve it properly. long story short: out of scope

igvm/hypervisor_preferences.py
Woellchen: `False` is an alias for `0.0` which makes the preference fail, yes

igvm/hypervisor_preferences.py
Woellchen: type hints do not make your code fail. they are just useful for static code analysis and documentation purposes

igvm/hypervisor_preferences.py
Woellchen: changed the comment. please check it again

igvm/hypervisor_preferences.py
Woellchen: fixed

igvm/hypervisor_preferences.py
Woellchen: true. fixed

igvm/hypervisor_preferences.py
Woellchen: done

igvm/hypervisor_preferences.py
Woellchen: done

igvm/hypervisor_preferences.py
Woellchen: done

igvm/hypervisor_preferences.py
Woellchen: done

igvm/hypervisor_preferences.py
```diff
@@ -1,63 +1,93 @@
...
+
+        @param vm: The VM object to check against the HV.
+        @param hv: The HV object to check against the VM.
+        @return: A float or a bool. The bool is just for convenience and will
+                 be converted to a float. A value of 0.0 means the HV does not
+                 fulfill the requirement at all. But a value of 1.0 means it is
+                 a very good candidate for this particular preference.
+        @rtype: Union[float, bool]
+        """
+        raise NotImplementedError('get_preference is not implemented')
```
lamaral, line 32: I think this log message should be `get_score` to reflect the function name

lamaral (changes requested):

igvm/hypervisor_preferences.py
```diff
@@ -1,63 +1,93 @@
...
             args += ', reserved=' + repr(self.reserved)
 
         return '{}({})'.format(type(self).__name__, args)
 
-    def __call__(self, vm, hv):
+    def get_score(self, vm, hv) -> Union[float, bool]:
         # Treat freshly created HVs always passing this check
         if not hv.dataset_obj[self.hv_attribute]:
-            return False
+            return True
```
kofrezo, line 61:
```suggestion
            # Because new installed Hypervisors don't have values 
            # (e.g. for load) - they are fresh meat ...
            return True
```

igvm/hypervisor_preferences.py
```diff
@@ -1,63 +1,93 @@
...
 
-        total_size = hv.dataset_obj[self.hv_attribute]
+        # Calculate the remaining "size" of the resource
+        total_size = hv.dataset_obj[self.hv_attribute] * self.multiplier
         vms_size = sum(
-            vm[self.vm_attribute] * self.multiplier
+            vm[self.vm_attribute]
             for vm in hv.dataset_obj['vms']
+            if vm['state'] != 'retired'
```
kofrezo, line 68: Does that work for all kind of resources ? If we assume retired VMs are shutdown (not guaranteed) and we are talking about CPUs and Memory the resources are not used but can we really always be certain a retired VM is shutdown and a resource is not used (I lack a good example but imagine we would calculate disk space this way).

kofrezo (changes requested):
2 minor suggestion for improvement. 1 question to clarify - overall great work!

TuxPowered42 (approved):

igvm/hypervisor_preferences.py
Woellchen: good catch

igvm/hypervisor_preferences.py
Woellchen: i get your point. first in our cases we do expect retired objects to not be in use anymore. you are right that it is not guaranteed but i don't see this as a real problem here because we do make later "hard" checks against libvirt before using a hypervisor. that way we are safe against all inconsistencies. these preferences just generate ordered hv "suggestions". whether a vm really fits on it is not the purpose here. is this fine for you?

igvm/hypervisor_preferences.py
Woellchen: i made the class `ABC` instead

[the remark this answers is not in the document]
Woellchen: i made the class `ABC` instead

igvm/hypervisor_preferences.py
kofrezo: Yes I think you are right this is safe enough and if it really fits is a second layer.

kofrezo (approved):

Woellchen:
retest this please

lamaral (approved):

PR #173, opened 2020-09-29
IGVM daemon protoype

kofrezo:


kofrezo:
This PR provides a prototype for a IGVM daemon by @HerrSpace to remove retire VMs. However the most important part we need right now is building and migrating which we decided to tackle in a simple approach where run IGVM in an endless loop and query Serveradmin for VMs to build based on a attribute. 

Lets not forget about this and pick up the ideas in next version of it.

PR #174, opened 2020-10-09
feat(tests): Add AWS as test environment for basic tests

lacunoc:
Adds basic test suite for AWS:
test_vm_build
test_delete
test_rebuild
test_start_stop
test_disk_set
test_sync
which is automatically using AWS environment, if the following env variables are configured:
VM_NET, 
AWS_ACCESS_KEY_ID, 
AWS_SECRET_ACCESS_KEY 
AWS_DEFAULT_REGION

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -794,30 +794,33 @@ def vm_rename(vm_hostname, new_hostname, offline=False):
     """
 
     with _get_vm(vm_hostname) as vm:
-        if vm.dataset_obj['datacenter_type'] != 'kvm.dct':
+        if vm.dataset_obj['datacenter_type'] in ['aws.dct', 'kvm.dct']:
+            if vm.dataset_obj['puppet_disabled']:
+                raise ConfigError(
+                    'Rename command only works with Puppet enabled'
+                )
+
+            if vm.dataset_obj['datacenter_type'] == 'kvm.dct':
+                _check_defined(vm)
+
+                if not offline:
+                    raise NotImplementedError(
+                        'Rename command only works with --offline at the moment.'
+                    )
+                if not vm.is_running():
+                    raise NotImplementedError(
+                        'Rename command only works online at the moment.'
+                    )
+
+                vm.rename(new_hostname)
+            elif vm.dataset_obj['datacenter_type'] == 'aws.dct':
+                vm.aws_rename(new_hostname)
+        else:
             raise NotImplementedError(
                 'This operation is not yet supported for {}'.format(
                     vm.dataset_obj['datacenter_type']
                 )
             )
-
-        if vm.dataset_obj['puppet_disabled']:
-            raise ConfigError(
-                'Rename command only works with Puppet enabled'
-            )
-
-        _check_defined(vm)
-
-        if not offline:
-            raise NotImplementedError(
-                'Rename command only works with --offline at the moment.'
-            )
-        if not vm.is_running():
-            raise NotImplementedError(
-                'Rename command only works online at the moment.'
-            )
-
-        vm.rename(new_hostname)
 
 
 @contextmanager
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -9,6 +9,7 @@ from collections import OrderedDict
 from contextlib import contextmanager, ExitStack
 from ipaddress import ip_address
 from os import environ
+from time import sleep
 
 from adminapi.dataset import Query
 from adminapi.filters import Any, StartsWith, Contains
@@ -304,6 +305,18 @@ def vm_build(vm_hostname, run_puppet=True, debug_puppet=False, postboot=None,
                 puppet_master=vm.dataset_obj['puppet_master'],
                 puppet_ca=vm.dataset_obj['puppet_ca'],
             )
+
+            if rebuild:
+                vm.aws_delete()
+                timeout_terminate = 60
+                instance_status = vm.aws_describe_instance_status(
+                    vm.dataset_obj['aws_instance_id'])
+                while (
+                    timeout_terminate and
+                    AWS_RETURN_CODES['terminated'] != instance_status
+                ):
+                    timeout_terminate -= 1
+                    sleep(1)
 
             vm.aws_build(
                 run_puppet=run_puppet,
@@ -794,33 +807,33 @@ def vm_rename(vm_hostname, new_hostname, offline=False):
     """
 
     with _get_vm(vm_hostname) as vm:
-        if vm.dataset_obj['datacenter_type'] in ['aws.dct', 'kvm.dct']:
-            if vm.dataset_obj['puppet_disabled']:
-                raise ConfigError(
-                    'Rename command only works with Puppet enabled'
-                )
-
-            if vm.dataset_obj['datacenter_type'] == 'kvm.dct':
-                _check_defined(vm)
-
-                if not offline:
-                    raise NotImplementedError(
-                        'Rename command only works with --offline at the moment.'
-                    )
-                if not vm.is_running():
-                    raise NotImplementedError(
-                        'Rename command only works online at the moment.'
-                    )
-
-                vm.rename(new_hostname)
-            elif vm.dataset_obj['datacenter_type'] == 'aws.dct':
-                vm.aws_rename(new_hostname)
-        else:
+        if vm.dataset_obj['datacenter_type'] not in ['aws.dct', 'kvm.dct']:
             raise NotImplementedError(
                 'This operation is not yet supported for {}'.format(
                     vm.dataset_obj['datacenter_type']
                 )
             )
+
+        if vm.dataset_obj['puppet_disabled']:
+            raise ConfigError(
+                'Rename command only works with Puppet enabled'
+            )
+
+        if vm.dataset_obj['datacenter_type'] == 'kvm.dct':
+            _check_defined(vm)
+
+            if not offline:
+                raise NotImplementedError(
+                    'Rename command only works with --offline at the moment.'
+                )
+            if not vm.is_running():
+                raise NotImplementedError(
+                    'Rename command only works online at the moment.'
+                )
+
+            vm.rename(new_hostname)
+        elif vm.dataset_obj['datacenter_type'] == 'aws.dct':
+            vm.aws_rename(new_hostname)
 
 
 @contextmanager
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -257,7 +257,7 @@ class VM(Host):
             current_state = (
                 response['StartingInstances'][0]['CurrentState']['Code']
             )
-            log.info(response)
+            log.debug(response)
             if current_state and current_state == AWS_RETURN_CODES['running']:
                 log.info('{} is already running.'.format(
                     self.dataset_obj['hostname']))
@@ -309,7 +309,7 @@ class VM(Host):
             )
             current_state = response['StoppingInstances'][0][
                 'CurrentState']['Code']
-            log.info(response)
+            log.debug(response)
             if current_state and current_state == AWS_RETURN_CODES['stopped']:
                 log.info('{} is already stopped.'.format(
                     self.dataset_obj['hostname']))
@@ -374,19 +374,22 @@ class VM(Host):
         try:
             response = ec2.terminate_instances(
                 InstanceIds=[self.dataset_obj['aws_instance_id']])
-            log.info(response)
+            log.debug(response)
         except ClientError as e:
             raise VMError(e)
 
     def is_running(self):
+        if self.dataset_obj['datacenter_type'] not in ['aws.dct', 'kvm.dct']:
+            raise NotImplementedError(
+                'This operation is not yet supported for {}'.format(
+                    self.dataset_obj['datacenter_type'])
+            )
         if self.dataset_obj['datacenter_type'] == 'kvm.dct':
             return self.hypervisor.vm_running(self)
-        elif self.dataset_obj['datacenter_type'] == 'aws.dct':
-            instance_status = self.aws_describe_instance_status(
-                self.dataset_obj['aws_instance_id'])
-            if instance_status == AWS_RETURN_CODES['running']:
-                return True
-            return False
+
+        instance_status = self.aws_describe_instance_status(
+            self.dataset_obj['aws_instance_id'])
+        return instance_status == AWS_RETURN_CODES['running']
 
     def wait_for_running(self, running=True, timeout=60):
         """
@@ -575,7 +578,7 @@ class VM(Host):
                     DryRun=False,
                     MinCount=1,
                     MaxCount=1)
-                log.info(response)
+                log.debug(response)
                 break
             except ClientError as e:
                 raise VMError(e)
@@ -660,7 +663,7 @@ class VM(Host):
 
             self.start(transaction=transaction)
 
-    def aws_rename(self, new_hostname) -> None:
+    def aws_rename(self, new_hostname: str) -> None:
         """AWS rename
 
         Rename a VM in AWS.
@@ -681,7 +684,7 @@ class VM(Host):
                 },
             ],
             DryRun=False)
-        log.info(response)
+        log.debug(response)
 
         self.run_puppet()
         self.aws_shutdown()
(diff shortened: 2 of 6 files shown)


igvm/commands.py
```diff
@@ -794,31 +794,34 @@ def vm_rename(vm_hostname, new_hostname, offline=False):
     """
 
     with _get_vm(vm_hostname) as vm:
-        if vm.dataset_obj['datacenter_type'] != 'kvm.dct':
+        if vm.dataset_obj['datacenter_type'] in ['aws.dct', 'kvm.dct']:
```
Woellchen, line 797: less is better. identations are bad. please return or raise as early as possible and skip all `else` statements if not really necessary. for example the first if should be negated to raise the error immediately: `if vm.dataset_obj['datacenter_type'] not in ['aws.dct', 'kvm.dct']:`. then leave the `else` away because the execution flow is already stopped.
it makes code so much more readable and easier to understand. whenever you introduce an unnecessary indentation you also put the burden on the "reader" of your code to build this structure in their heads in order to understand it. debugging also becomes much harder.

igvm/vm.py
```diff
@@ -378,7 +379,14 @@ def aws_delete(self):
             raise VMError(e)
 
     def is_running(self):
-        return self.hypervisor.vm_running(self)
+        if self.dataset_obj['datacenter_type'] == 'kvm.dct':
+            return self.hypervisor.vm_running(self)
+        elif self.dataset_obj['datacenter_type'] == 'aws.dct':
+            instance_status = self.aws_describe_instance_status(
+                self.dataset_obj['aws_instance_id'])
+            if instance_status == AWS_RETURN_CODES['running']:
```
Woellchen, line 387: there is no need for an `if` at all here. just `return instance_status == AWS_RETURN_CODES['running']`

igvm/vm.py
```diff
@@ -378,7 +379,14 @@ def aws_delete(self):
             raise VMError(e)
 
     def is_running(self):
-        return self.hypervisor.vm_running(self)
+        if self.dataset_obj['datacenter_type'] == 'kvm.dct':
+            return self.hypervisor.vm_running(self)
+        elif self.dataset_obj['datacenter_type'] == 'aws.dct':
```
Woellchen, line 384: no need for an `elif` here. you are already returning above. if you want to make sure the `datacenter_type` is actually supported then you could check it right in the beginning of the function and otherwise raise an error.

igvm/vm.py
```diff
@@ -652,6 +660,33 @@ def rename(self, new_hostname):
...
+        response = ec2.create_tags(
+            Resources=[self.dataset_obj['aws_instance_id']],
+            Tags=[
+                {
+                    'Key': 'Name',
+                    'Value': new_hostname,
+                },
+            ],
+            DryRun=False)
+        log.info(response)
```
Woellchen, line 684: this looks like a `debug` message to me

tests/conftest.py
```diff
@@ -57,6 +59,9 @@ def clean_all(route_network, vm_hostname=None):
     for hv in hvs:
         clean_hv(hv, pattern)
```
Woellchen, line 60: if `datacenter_type` is `aws.dct` do we actually need to clean hvs then?

tests/conftest.py
```diff
@@ -115,18 +120,37 @@ def clean_serveradmin(filters):
     Query(filters).delete().commit()
 
 
+def clean_aws(vm_hostname):
+    try:
+        vm_stop(vm_hostname)
```
Woellchen, line 125: the purpose of such tests is to test functions such as `vm_stop` and `vm_delete` and not to use them when not actually testing them. we can never rely on them because they can be broken anytime. the cleanup must be done "manually" without using any existing code. otherwise we might run into a hen-egg problem. please add some code here to hard-kill those vms on aws

tests/conftest.py
```diff
@@ -115,18 +120,37 @@ def clean_serveradmin(filters):
...
-
-    return list(subnets)[PYTEST_XDIST_WORKER][index]
+    if len(non_vm_hosts) > subnets[0].num_addresses:
+        raise Exception(
+            'Can\'t split {} into enough subnets '
+            'for {} parallel tests'.format(
+                vm_net, PYTEST_XDIST_WORKER_COUNT,
+            )
+        )
+    return list(subnets)[PYTEST_XDIST_WORKER + 1][index]
```
Woellchen, line 156: i know i suggested this to you in the first place but we don't actually need the `list()` here again. it's already done above when assigning it to the `subnets` variable

tests/test_integration.py
```diff
@@ -221,8 +247,12 @@ def test_delete(self):
         with self.assertRaises(IGVMError):
             vm_delete(VM_HOSTNAME)
 
-        with _get_vm(VM_HOSTNAME) as vm:
-            vm.shutdown()
+        if self.datacenter_type == 'kvm.dct':
+            with _get_vm(VM_HOSTNAME) as vm:
+                vm.shutdown()
+        elif self.datacenter_type == 'aws.dct':
+            vm_stop(VM_HOSTNAME)
```
Woellchen, line 254: i am not a fan of this. now we have two different functions from two different places to stop a vm? `vm.shutdown()` should work for both kvm and aws. this `vm.shutdown()` should internally decide what to do based on the `datacenter_type`. we should always aim for a clear api that is consistent

tests/test_integration.py
```diff
@@ -249,9 +279,28 @@ def test_rebuild(self):
...
 
-        # Now stop it and rebuild it
         vm_stop(VM_HOSTNAME)
-        vm_build(VM_HOSTNAME, rebuild=True)
+
+        if self.datacenter_type == 'kvm.dct':
+            # Now stop it and rebuild it
+            vm_build(VM_HOSTNAME, rebuild=True)
+        elif self.datacenter_type == 'aws.dct':
+            vm_delete(VM_HOSTNAME, retire=True)
```
Woellchen, line 288: same here. we should try to develop clear and consistent apis. i would expect `vm_build(VM_HOSTNAME, rebuild=True)` to do exactly the same thing on aws that it would do in our own datacenter. the `vm_build()` function should internally decide what to do based on the `datacenter_type` and not the test :)

tests/test_integration.py
```diff
@@ -293,49 +342,67 @@ def test_start_stop(self):
...
 
-        size = 8
-        disk_set(VM_HOSTNAME, '{}GB'.format(size))
-        obj = Query({'hostname': VM_HOSTNAME}, ['disk_size_gib']).get()
-
-        self.assertEqual(obj['disk_size_gib'], size)
-        self.assertEqual(_get_disk_hv(), size)
-        self.assertEqual(_get_disk_vm(), size)
+        if self.datacenter_type == 'kvm.dct':
+            disk_set(VM_HOSTNAME, '{}GB'.format(size))
```
Woellchen, line 384: is there no disk increasing for aws?!

tests/test_integration.py
```diff
@@ -632,7 +699,7 @@ def tearDown(self):
             vm['hostname'] = RenameTest._get_renamed_hostname(vm['hostname'])
 
         clean_cert(vm)
-        clean_all(vm['hostname'])
+        clean_all(self.datacenter_type, vm['hostname'])
```
Woellchen, line 702: hm i wonder how this is even supposed to work without the `self.route_network`. but it was like that before so i guess we can ignore this for now

Woellchen (changes requested):

igvm/vm.py
```diff
@@ -652,6 +660,33 @@ def rename(self, new_hostname):
 
             self.start(transaction=transaction)
 
+    def aws_rename(self, new_hostname) -> None:
```
kofrezo, line 663: Why no type hint for self and new_hostname ?

tests/__init__.py
```diff
@@ -24,8 +24,10 @@
 else:
     PYTEST_XDIST_WORKER_COUNT = 1
 
-
-VM_NET = 'igvm-net-{}-aw.test.ig.local'.format(JENKINS_EXECUTOR)
+if environ.get('VM_NET'):
+    VM_NET = environ['VM_NET']
+else:
+    VM_NET = 'igvm-net-{}-aw.test.ig.local'.format(JENKINS_EXECUTOR)
```
kofrezo, line 30: You could write 
```
environ.net('VM_NET', default='igvm-net-{}-aw.test.ig.local'.format(JENKINS_EXECUTOR))
```

tests/conftest.py
```diff
@@ -115,18 +120,37 @@ def clean_serveradmin(filters):
...
         raise Exception(
             'Can\'t split {} into enough subnets '
             'for {} parallel tests'.format(
                 vm_net, PYTEST_XDIST_WORKER_COUNT,
             )
         )
-
-    return list(subnets)[PYTEST_XDIST_WORKER][index]
+    if len(non_vm_hosts) > subnets[0].num_addresses:
+        raise Exception(
```
kofrezo, line 150: Please don't raise Exception but rather use a existing one or create a new one such as e.g. `IGVMTestError`

tests/test_integration.py
kofrezo: I strongly aggree on this one

tests/test_integration.py
```diff
@@ -293,49 +342,67 @@ def test_start_stop(self):
...
+                    .strip(),
+                    'G'
+                )
+        elif self.datacenter_type == 'aws.dct':
+            def _get_disk_vm():
+                partition = self.vm.run('findmnt -nro SOURCE /')
+                disk = self.vm.run('lsblk -nro PKNAME {}'.format(partition))
+                disk_size = self.vm.run(
+                    'lsblk -bdnro size /dev/{}'.format(disk))
+                disk_size_gib = int(disk_size) / 1024 / 1024 / 1024
```
kofrezo, line 364: How about 
```
disk_size_gib = int(disk_size) / 1024**3
```

tests/test_integration.py
kofrezo: It was broken before. It used `vm['hostname']` as filter for the route_network to query all Hypervisors and did not find any so did nothing. This might actually be a spot where we are not properly cleaning up things so lets fix it.

kofrezo (changes requested):
Overall great work. Just some code or style issues and I strongly agree with @Woellchen that actions such as shutdown should exist once and it should be transparent for the user whatever the VM is running in AWS or in our datacenter with KVM.

igvm/commands.py
lacunoc: error is now raised in the beginning

igvm/vm.py
lacunoc: changed to debug

tests/test_integration.py
lacunoc: In my opinion, it was wrong in the first place to use `vm.shutdown()`, since we usually call the commands, which we progragate, so I reworked it, so both datacenter_types are using their shared `vm_stop` command.

tests/test_integration.py
lacunoc: reworked

tests/test_integration.py
lacunoc: There is, but only one time, because then the disk is in optimizing state for unknown amount of time. I didn't want it to fail randomly.

tests/test_integration.py
lacunoc: fixed

igvm/vm.py
lacunoc: added

tests/__init__.py
lacunoc: changed

tests/test_integration.py
lacunoc: changed

tests/conftest.py
lacunoc: Added `IGVMTestError`

igvm/vm.py
lacunoc: changed

tests/conftest.py
lacunoc: No, we don't. Added `kvm.dct` here.

tests/test_integration.py
Woellchen: but is the unknown amount of time seconds or minutes or hours? i would find it better to test this still. if it really takes up to hours then we could still make the call and check for a valid response no?

tests/test_integration.py
Woellchen: yup i think it's good. it is consistent now with the rest of the code and we do not have business logic in the test :)

igvm/vm.py
lacunoc: reworked this function

tests/test_integration.py
lacunoc: It is unknown for at least minutes, depending on the size of disk and increase. I'm checking the state of the volume now and if this is in state `optimizing` the disk_set raises an Exception, which I also check on the integration tests.

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

tests/conftest.py
lacunoc: I reworked the cleanup for AWS, which should be fine now.

tests/conftest.py
```diff
@@ -115,18 +124,77 @@ def clean_serveradmin(filters):
...
+                    ]
+                },
+            ],
+            InstanceIds=[obj['aws_instance_id']],
+            DryRun=False)['Reservations'][0]['Instances'][0]['State']['Code']
+
+        return int(response)
+
+    try:
+        obj = Query({'hostname': vm_hostname}, ['aws_instance_id']).get()
```
Woellchen, line 149: does a missing serveradmin really mean there is NO possibility a machine is running on aws?

tests/conftest.py
```diff
@@ -115,18 +124,77 @@ def clean_serveradmin(filters):
...
+    except DatasetError:  # No object to clean up
+        return
+
+    if obj['aws_instance_id']:
```
Woellchen, line 153: please try to NOT indent your code. please negate this and return early

tests/conftest.py
```diff
@@ -115,18 +124,77 @@ def clean_serveradmin(filters):
...
+        timeout = 120
+        ec2 = boto3.client('ec2')
+        try:
+            ec2.stop_instances(
+                InstanceIds=[obj['aws_instance_id']], DryRun=False
+            )
+        except ClientError as e:
+            pass  # Not running
+        for retry in range(timeout):
```
Woellchen, line 162: you are not using the `retry` variable. use `for _ in range(timeout)`

tests/conftest.py
```diff
@@ -115,18 +124,77 @@ def clean_serveradmin(filters):
...
+            instance_status = _get_instance_status()
+            if AWS_RETURN_CODES['stopped'] == instance_status:
+                break
+            sleep(1)
+
+        ec2.terminate_instances(InstanceIds=[obj['aws_instance_id']])
+        for retry in range(timeout):
```
Woellchen, line 169: same here with `_`

tests/conftest.py
```diff
@@ -115,18 +124,77 @@ def clean_serveradmin(filters):
...
+            instance_status = _get_instance_status()
+            if AWS_RETURN_CODES['terminated'] == instance_status:
```
Woellchen, line 171: what happens if it isnt terminated in 120s?

Woellchen (approved):

kofrezo (approved):

tests/conftest.py
lacunoc: changed

tests/conftest.py
lacunoc: changed

tests/conftest.py
lacunoc: changed

tests/conftest.py
lacunoc: Usually a machine is terminate within seconds, so 120 should be more than enough.

tests/conftest.py
lacunoc: At least for the tests this is the case. You shouldn't spawn any VMs on AWS side without using igvm and serveradmin

PR #176, opened 2020-11-10
Add xfs offline transport

Felixoid:
- Add xfsdump/xfsrestore offline migration transport
- Add disk changing during xfs offline migrations
- Add tests

### DIFF
--- a/igvm/cli.py
+++ b/igvm/cli.py
@@ -165,8 +165,10 @@ def parse_args():
     subparser.add_argument(
         '--offline-transport',
         default='drbd',
+        choices=('drbd', 'netcat', 'xfs'),
         help=(
-            'Specify drbd (default) or netcat transport to migrate disk image'
+            'Specify drbd (default), netcat or xfs transport to migrate '
+            'disk image'
         ),
     )
     subparser.add_argument(
--- a/igvm/cli.py
+++ b/igvm/cli.py
@@ -185,6 +185,13 @@ def parse_args():
         action='store_true',
         help='Build or migrate VM only to a HV with the same environment of VM'
     )
+    subparser.add_argument(
+        '--disk-size',
+        dest='disk_size',
+        type=int,
+        help='Resize disk of migrated VM. '
+        'Works only with --offline --offline-transport=xfs',
+    )
 
     subparser = subparsers.add_parser(
         'change-address',
--- a/igvm/cli.py
+++ b/igvm/cli.py
@@ -174,7 +174,7 @@ def parse_args():
         action='store_true',
         help=(
             'Don\'t shutdown VM during offline migration, igvm will wait for'
-            ' operator to shut down VM.'
+            ' operator to shut down VM for 24h.'
         ),
     )
     subparser.add_argument(
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -368,7 +368,7 @@ def vm_migrate(vm_hostname=None, vm_object=None, hypervisor_hostname=None,
                run_puppet=False, debug_puppet=False,
                offline=False, offline_transport='drbd',
                allow_reserved_hv=False, no_shutdown=False,
-               enforce_vm_env=False):
+               enforce_vm_env=False, disk_size=None):
     """Migrate a VM to a new hypervisor."""
 
     if not (bool(vm_hostname) ^ bool(vm_object)):
@@ -391,6 +391,13 @@ def vm_migrate(vm_hostname=None, vm_object=None, hypervisor_hostname=None,
                     _vm.dataset_obj['datacenter_type'])
             )
 
+        # We have to check migration settings before searching for a HV,
+        # because the new disk size must be checked and set
+        current_size_gib = _vm.dataset_obj['disk_size_gib']
+        _vm.dataset_obj['disk_size_gib'] = _vm.hypervisor.vm_new_disk_size(
+            _vm, offline, offline_transport, disk_size
+        )
+
         if hypervisor_hostname:
             hypervisor = es.enter_context(_get_hypervisor(
                 hypervisor_hostname, allow_reserved=allow_reserved_hv
@@ -407,6 +414,10 @@ def vm_migrate(vm_hostname=None, vm_object=None, hypervisor_hostname=None,
                 offline,
                 enforce_vm_env,
             ))
+
+        # After the HV is chosen, disk_size_gib must be restored
+        # to pass _check_attributes(_vm)
+        _vm.dataset_obj['disk_size_gib'] = current_size_gib
 
         was_running = _vm.is_running()
 
@@ -429,7 +440,7 @@ def vm_migrate(vm_hostname=None, vm_object=None, hypervisor_hostname=None,
         with Transaction() as transaction:
             _vm.hypervisor.migrate_vm(
                 _vm, hypervisor, offline, offline_transport, transaction,
-                no_shutdown,
+                no_shutdown, disk_size,
             )
 
             previous_hypervisor = _vm.hypervisor
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -368,7 +368,10 @@ class Hypervisor(Host):
         if new_size_gib < vm.dataset_obj['disk_size_gib']:
             raise NotImplementedError(
                 'Cannot shrink the disk. '
-                'Use `migrate --offline --offline-transport xfs --disk-size X`'
+                'Use `igvm migrate --offline --offline-transport xfs '
+                '--disk-size {} {}`'.format(
+                    new_size_gib, vm.fqdn,
+                )
             )
         volume = self.get_volume_by_vm(vm)
         if self.get_storage_type() == 'logical':
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -18,7 +18,9 @@ from igvm.exceptions import (
     HypervisorError,
     InconsistentAttributeError,
     InvalidStateError,
+    RemoteCommandError,
     StorageError,
+    XfsMigrationError,
 )
 from igvm.host import Host
 from igvm.kvm import (
@@ -583,7 +585,7 @@ class Hypervisor(Host):
         self, vm: VM, target_hypervisor: 'Hypervisor', offline: bool,
         offline_transport: str, transaction: Transaction, no_shutdown: bool,
     ):
-        if offline_transport not in ['netcat', 'drbd']:
+        if offline_transport not in ['netcat', 'drbd', 'xfs']:
             raise StorageError(
                 'Unknown offline transport method {}!'
                 .format(offline_transport)
@@ -641,6 +643,15 @@ class Hypervisor(Host):
                         vm.dataset_obj['disk_size_gib'] * 1024 ** 3,
                         args,
                     )
+            elif offline_transport == 'xfs':
+                self._wait_for_shutdown(vm, no_shutdown, transaction)
+                with target_hypervisor.xfsrestore(vm, transaction) as listener:
+                    self.xfsdump(vm, listener, transaction)
+
+                target_hypervisor.wait_for_xfsrestore(vm)
+                target_hypervisor.check_xfsrestore_log(vm)
+                target_hypervisor.umount_vm_storage(vm)
+
             target_hypervisor.define_vm(vm, transaction)
         else:
             # For online migrations always use same volume name as VM
@@ -860,6 +871,87 @@ class Hypervisor(Host):
             '| /bin/nc.openbsd -q 1 {2} {3}'
             .format(device, size, listener[0], listener[1])
         )
+
+    def _xfsrestore_log_name(self, vm: VM) -> str:
+        device_name = self.get_volume_by_vm(vm).name()
+        return '/tmp/xfsrestore-{}.log'.format(device_name)
+
+    def check_xfsrestore_log(self, vm: VM):
+        """
+        Search for WARNING in the xfsrestore log file.
+        Raises exception if found
+        """
+        self.run('cat {}'.format(self._xfsrestore_log_name(vm)))
+        try:
+            self.run('grep -q WARNING {} && exit 1 || exit 0'
+                     .format(self._xfsrestore_log_name(vm)))
+        except RemoteCommandError:
+            raise XfsMigrationError('xfs dump/restore caused warnings')
+
+    @contextmanager
+    def xfsrestore(
+        self, vm: VM, transaction: Transaction = None,
+    ) -> Iterator[Tuple[str, int]]:
+        """
+        Formats a vm's storage, mounts it, spawns background netcat process
+        and pipes the load to xfsrestore command to restore xfsdump on the
+        target HV
+
+        :param VM vm: The migrating VM
+        :param Transaction transaction: The transaction to rollback
+        :rtype Iterator[Tuple[str, int]]: (fqdn, port) pair
+        """
+        device = self.get_volume_by_vm(vm).path()
+        port = self._netcat_port(device)
+        mount_dir = self.format_vm_storage(vm, transaction)
+
+        # xfsrestore args:
+        #  -F: Don't prompt the operator.
+        #  -J: inhibits inventory update
+        # xfsrestore needs to output its logs, otherwise it fails
+        self.run(
+            'nohup /bin/nc.openbsd -l -p {0} '
+            '| xfsrestore -F -J - {1} 2>{2} 1>&2 &'
+            .format(port, mount_dir, self._xfsrestore_log_name(vm)),
+            pty=False,  # Has to be here for background processes
+        )
+        try:
+            yield self.fqdn, port
+        except BaseException:
+            self.kill_netcat(port)
+            raise
+
+    def wait_for_xfsrestore(self, vm: VM):
+        """
+        On the HV with slow IO the process must wait until xfsrestore is done
+        """
+        mount_dir = self.vm_mount_path(vm)
+        self.run(
+            'while pgrep -f "^xfsrestore -F -J - {0}"; do sleep 1; done'
+            .format(mount_dir)
+        )
+
+    def xfsdump(self, vm: VM, listener, transaction: Transaction = None):
+        """
+        Mounts the vm's storage, and then dumps the device via xfsdump and
+        netcat to a remote listener
+
+        :param VM vm: The migrating VM
+        :param Transaction transaction: The transaction to rollback
+        """
+        mount_dir = self.mount_vm_storage(vm, transaction)
+
+        # xfsdump args:
+        #  -l: level 0 is an absolute dump
+        #  -F: Don't prompt the operator.
+        #  -J: inhibits inventory update
+        #  -p: progress update interval
+        self.run(
+            'xfsdump -o -l 0 -F -J -p 1 - {0} '
+            '| /bin/nc.openbsd -q 1 {1} {2}'
+            .format(mount_dir, listener[0], listener[1]),
+        )
+        self.umount_vm_storage(vm)
 
     def estimate_cpu_cores_used(self, vm: VM) -> float:
         """Estimate the number of CPU cores used by the VM
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -366,7 +366,10 @@ class Hypervisor(Host):
     def vm_set_disk_size_gib(self, vm, new_size_gib):
         """Changes disk size of a VM."""
         if new_size_gib < vm.dataset_obj['disk_size_gib']:
-            raise NotImplementedError('Cannot shrink the disk.')
+            raise NotImplementedError(
+                'Cannot shrink the disk. '
+                'Use `migrate --offline --offline-transport xfs --disk-size X`'
+            )
         volume = self.get_volume_by_vm(vm)
         if self.get_storage_type() == 'logical':
             # There is no resize function in version of libvirt
@@ -559,6 +562,71 @@ class Hypervisor(Host):
         the guest OS"""
         return 'vda1'
 
+    def check_migrate_parameters(
+        self, vm: VM, offline: bool, offline_transport: str,
+        disk_size: int = None,
+    ):
+        if offline_transport not in ['netcat', 'drbd', 'xfs']:
+            raise StorageError(
+                'Unknown offline transport method {}!'
+                .format(offline_transport)
+            )
+
+        if disk_size is None:
+            return
+
+        if disk_size < 1:
+            raise StorageError('disk_size must be at least 1GiB!')
+        if not (offline and offline_transport == 'xfs'):
+            raise StorageError(
+                'disk_size can be applied only with offline transport xfs!'
+            )
+        allocated_space = vm.dataset_obj['disk_size_gib'] - vm.disk_free()
+        if disk_size < allocated_space:
+            raise StorageError(
+                'disk_size is lower than allocated space: {} < {}!'
+                .format(disk_size, allocated_space)
+            )
+
+    def vm_new_disk_size(
+        self, vm: VM, offline: bool, offline_transport: str,
+        disk_size: int = None,
+    ) -> int:
+        self.check_migrate_parameters(
+            vm, offline, offline_transport, disk_size
+        )
+        if disk_size is None:
+            return vm.dataset_obj['disk_size_gib']
+        return disk_size or vm.dataset_obj['disk_size_gib']
+
+    def _vm_apply_new_disk_size(
+        self, vm: VM, offline: bool, offline_transport: str,
+        transaction: Transaction, disk_size: int = 0,
+    ):
+        """
+        If the new VM disk size is set, checks if it's correct and sufficient
+        and commit the new size. Rolls it back on the interrupted migration
+
+        :param VM vm: The migrating VM
+        :param str offline_transport: offline migration transport
+        :param Transaction transaction: The transaction to rollback
+        :param int disk_size: the new disk_size_gib attribute
+        """
+        size = self.vm_new_disk_size(vm, offline,
+                                     offline_transport, disk_size)
+        if size == vm.dataset_obj['disk_size_gib']:
+            return
+
+        old_size = vm.dataset_obj['disk_size_gib']
+        vm.dataset_obj['disk_size_gib'] = size
+        vm.dataset_obj.commit()
+
+        if transaction:
+            def restore_size():
+                vm.dataset_obj['disk_size_gib'] = old_size
+                vm.dataset_obj.commit()
+            transaction.on_rollback('reset_disk_size', restore_size)
+
     def _wait_for_shutdown(
         self, vm: VM, no_shutdown: bool, transaction: Transaction,
     ):
@@ -584,12 +652,11 @@ class Hypervisor(Host):
     def migrate_vm(
         self, vm: VM, target_hypervisor: 'Hypervisor', offline: bool,
         offline_transport: str, transaction: Transaction, no_shutdown: bool,
+        disk_size: int = 0,
     ):
-        if offline_transport not in ['netcat', 'drbd', 'xfs']:
-            raise StorageError(
-                'Unknown offline transport method {}!'
-                .format(offline_transport)
-            )
+        self._vm_apply_new_disk_size(
+            vm, offline, offline_transport, transaction, disk_size
+        )
 
         if offline:
             log.info(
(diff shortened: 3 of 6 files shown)


igvm/hypervisor.py
```diff
@@ -828,6 +835,48 @@ def device_to_netcat(self, device, size, listener):
...
 
+    @contextmanager
+    def xfsrestore(self, vm: VM, transaction: Transaction = None):
+        device = self.get_volume_by_vm(vm).path()
+        port = self.netcat_port(device)
+        mount_dir = self.format_vm_storage(vm, transaction)
+
+        # Check xfsrestore exists, since problem is hidden with
+        # 'nohup', '&' and 'pty=False'
+        self.run('xfsrestore --help', silent=True)
```
TuxPowered42, line 894: We are never checking for any other programs as we expect them to be deployed by Puppet. Why do we need to check for this one? And if we go with the way of checking, we should add even more checks, preferrably in another PR, there was even a ticket somewhere to check if it is possible to SSH to a VM before igvm starts, that seems a good idea to implemment too.

igvm/hypervisor.py
```diff
@@ -828,6 +835,48 @@ def device_to_netcat(self, device, size, listener):
             .format(device, size, listener[0], listener[1])
         )
 
+    @contextmanager
+    def xfsrestore(self, vm: VM, transaction: Transaction = None):
+        device = self.get_volume_by_vm(vm).path()
+        port = self.netcat_port(device)
+        mount_dir = self.format_vm_storage(vm, transaction)
+
+        # Check xfsrestore exists, since problem is hidden with
```
TuxPowered42, line 892: "if ... exists"

igvm/hypervisor.py
```diff
@@ -828,6 +835,48 @@ def device_to_netcat(self, device, size, listener):
...
+        )
+        try:
+            yield self.fqdn, port
+        except BaseException:
+            self.kill_netcat(port)
+            raise
+
+    def xfsdump(self, vm: VM, listener, transaction: Transaction = None):
+        mount_dir = self.mount_vm_storage(vm, transaction)
+        # Check xfsdump exists, since problem is hidden with pipe
```
TuxPowered42, line 914: "if ... exists"

TuxPowered42 (changes requested):

igvm/hypervisor.py
```diff
@@ -567,37 +566,45 @@ def migrate_vm(
                 .format(offline_transport)
             )
         disk_size = disk_size if 0 < disk_size else 0
-        if offline_transport != 'xfs' and 0 < disk_size:
+        if not (offline and offline_transport == 'xfs') and 0 < disk_size:
```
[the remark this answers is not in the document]
Felixoid: I'm trying to find an article where I've read about the positive side of using `<` always.

If shortly, because it uses the "natural" x axis `------->`, it allows decreasing the brain load to parse the conditions.

I'm doing my best to find the article. =\

igvm/cli.py
```diff
@@ -165,8 +165,10 @@ def parse_args():
     subparser.add_argument(
         '--offline-transport',
         default='drbd',
+        choices=('drbd', 'netcat', 'xfs'),
         help=(
-            'Specify drbd (default) or netcat transport to migrate disk image'
+            'Specify drbd (default), netcat or xfs transport to migrate '
```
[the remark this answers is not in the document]
Felixoid: I'll add the formatter, because I like the idea of displaying defaults.

igvm/hypervisor.py
```diff
@@ -606,35 +621,15 @@ def migrate_vm(
...
-                            transaction=transaction,
-                        )
-
+                self.migrate_wait_shutdown(vm, no_shutdown, transaction)
                 vm_disk_path = target_hypervisor.get_volume_by_vm(vm).path()
                 with target_hypervisor.netcat_to_device(vm_disk_path) as args:
                     self.device_to_netcat(
                         self.get_volume_by_vm(vm).path(),
-                        vm.dataset_obj['disk_size_gib'] * 1024 ** 3,
+                        vm.dataset_obj['disk_size_gib'] * (1 << 30),
```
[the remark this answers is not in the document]
Felixoid: at least that's how it's used everywhere in our puppet. From the preferences, I like more the bit notation.

igvm/cli.py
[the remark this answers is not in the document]
Felixoid: Changed my mind. It's out of the current scope since I'm not able to follow the logic of `IGVMArgumentParser` quickly.

igvm/hypervisor.py
```diff
@@ -559,12 +559,24 @@ def vm_block_device_name(self):
...
         offline_transport: str, transaction: Transaction, no_shutdown: bool,
+        disk_size: int = 0,
     ):
         if offline_transport not in ['netcat', 'drbd', 'xfs']:
             raise StorageError(
                 'Unknown offline transport method {}!'
                 .format(offline_transport)
             )
+        disk_size = disk_size if 0 < disk_size else 0
+        if offline_transport != 'xfs' and 0 < disk_size:
```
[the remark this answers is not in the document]
Felixoid: Here I've found a hint why is it better https://flylib.com/books/en/2.823.1.155/1/, Search on the page by `Writing Numeric Expressions in Number-Line Order`

igvm/hypervisor.py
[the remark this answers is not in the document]
Felixoid: Here I've found a hint why is it better https://flylib.com/books/en/2.823.1.155/1/, Search on the page by `Writing Numeric Expressions in Number-Line Order`

Felixoid:
retest it

igvm/hypervisor.py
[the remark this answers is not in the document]
kofrezo: I agree with @TuxPowered42 this is way less readable please keep the current way of writing.

igvm/hypervisor.py
```diff
@@ -802,18 +797,22 @@ def check_netcat(self, port):
     def kill_netcat(self, port):
         self.run('pkill -f "^/bin/nc.openbsd -l -p {}"'.format(port))
 
-    @contextmanager
-    def netcat_to_device(self, device):
+    def netcat_port(self, device):
```
kofrezo, line 855: I think it is a good idea to split this into a dedicated method :+1:. 

It might be also worth to make it a underscore method `_netcat_port` to hint users that this is a internal method that should not get used outside of the module.

igvm/hypervisor.py
```diff
@@ -802,18 +797,22 @@ def check_netcat(self, port):
...
         dev_minor = self.run('stat -L -c "%T" {}'.format(device), silent=True)
         dev_minor = int(dev_minor, 16)
         port = 7000 + dev_minor
-
         self.check_netcat(port)
+        return port
+
+    @contextmanager
+    def netcat_to_device(self, device):
```
kofrezo, line 863: A doc string would be helpful here.

igvm/hypervisor.py
```diff
@@ -555,9 +556,23 @@ def vm_block_device_name(self):
         the guest OS"""
         return 'vda1'
 
+    def migrate_wait_shutdown(
+        self, vm: VM, no_shutdown: bool, transaction: Transaction,
+    ):
```
kofrezo, line 610: This should method should be a parameter of `migrate_vm`.

igvm/hypervisor.py
```diff
@@ -555,9 +556,23 @@ def vm_block_device_name(self):
...
+                vm.shutdown(
+                    check_vm_up_on_transaction=False,
+                    transaction=transaction,
+                )
+
     def migrate_vm(
-        self, vm, target_hypervisor, offline, offline_transport, transaction,
-        no_shutdown,
+        self, vm: VM, target_hypervisor: 'Hypervisor', offline: bool,
+        offline_transport: str, transaction: Transaction, no_shutdown: bool,
```
kofrezo, line 624: You can make adding type hints a dedicated commit the next time to make it easier for people to distinct style changes from actual changes.

igvm/cli.py
```diff
@@ -185,6 +185,14 @@ def parse_args():
         action='store_true',
         help='Build or migrate VM only to a HV with the same environment of VM'
     )
+    subparser.add_argument(
+        '--disk-size',
+        dest='disk_size',
+        default=0,
```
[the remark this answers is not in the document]
kofrezo: If it is a required argument to the command and and should not be possible to run it without specifying a disk size you can make it a positional argument.

igvm/commands.py
```diff
@@ -391,6 +391,13 @@ def vm_migrate(vm_hostname=None, vm_object=None, hypervisor_hostname=None,
                     _vm.dataset_obj['datacenter_type'])
             )
 
+        # We have to check migration settings before searching for a HV,
+        # because the new disk size must be checked and set
+        current_size_gib = _vm.dataset_obj['disk_size_gib']
+        _vm.dataset_obj['disk_size_gib'] = _vm.hypervisor.migrate_disk_size(
+            _vm, offline, offline_transport, disk_size
+        )
```
kofrezo, line 399: Does this also work for the algorithm to automatically choose a Hypervisor ?

igvm/hypervisor.py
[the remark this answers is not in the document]
kofrezo: I agree with @TuxPowered42. 

Even if your article has an interesting point in here we should stick to what we use more often in the whole project and not change the style in at some places in the project.

Please take this into account for further comparisons too.

igvm/hypervisor.py
```diff
@@ -555,15 +556,77 @@ def vm_block_device_name(self):
...
+        if 0 < disk_size < allocated_space:
+            raise StorageError(
+                'disk_size is lower than allocated space: {} < {}!'
+                .format(disk_size, allocated_space)
+            )
+
+    def migrate_disk_size(
+        self, vm: VM, offline: bool, offline_transport: str,
+        disk_size: int = 0,
+    ) -> int:
```
kofrezo, line 583: I was confused in the first place when reading the method name because I thought it would actually "migrate" or change the disk size. Maybe we can find a better name for this by inverting like e.g. `disk_size_for_migration`

igvm/hypervisor.py
```diff
@@ -555,15 +556,77 @@ def vm_block_device_name(self):
...
+        self.check_migrate_vm(vm, offline, offline_transport, disk_size)
+        disk_size = disk_size if 0 < disk_size else 0
+        return disk_size or vm.dataset_obj['disk_size_gib']
+
+    def migrate_change_disk_size(
+        self, vm: VM, offline: bool, offline_transport: str,
+        transaction, disk_size: int = 0,
+    ):
```
kofrezo, line 591: Same here. The name does not really say what the method is doing IMHO. Maybe we use something like `apply_new_disk_size` or `save_new_disk_size` and a doc string.

igvm/hypervisor.py
```diff
@@ -555,15 +556,77 @@ def vm_block_device_name(self):
...
+                vm.dataset_obj.commit()
+            transaction.on_rollback('reset_disk_size', restore_size)
+
+
+    def migrate_wait_shutdown(
+        self, vm: VM, no_shutdown: bool, transaction: Transaction,
+    ):
+        vm.set_state('maintenance', transaction=transaction)
+        if vm.is_running():
+            if no_shutdown:
```
kofrezo, line 613: I was not expecting that - maybe `wait_for_shutdown` or `manual_shutdown` is a better name.

igvm/hypervisor.py
```diff
@@ -555,15 +556,77 @@ def vm_block_device_name(self):
...
+                log.info('Please shut down the VM manually now')
+                vm.wait_for_running(running=False, timeout=86400)
```
kofrezo, line 615: I don't think it is realistic that we have to wait 24 hours to shutdown a VM. I suggest using a smaller timeout such as 5 minutes and check again if the VM was shut down and if not raise an Exception.

igvm/hypervisor.py
```diff
@@ -826,8 +880,50 @@ def device_to_netcat(self, device, size, listener):
         self.run(
             'dd if={0} ibs=1048576 | pv -f -s {1} '
             '| /bin/nc.openbsd -q 1 {2} {3}'
-            .format(device, size, *listener)
+            .format(device, size, listener[0], listener[1])
+        )
+
+    @contextmanager
+    def xfsrestore(self, vm: VM, transaction: Transaction = None):
```
kofrezo, line 887: This a quite powerful and new method please add a doc string.

igvm/hypervisor.py
```diff
@@ -826,8 +880,50 @@ def device_to_netcat(self, device, size, listener):
...
+            .format(port, mount_dir),
+            pty=False,  # Has to be here for background processes
+        )
+        try:
+            yield self.fqdn, port
+        except BaseException:
+            self.kill_netcat(port)
+            raise
+
+    def xfsdump(self, vm: VM, listener, transaction: Transaction = None):
```
kofrezo, line 912: Same here. Please add a doc string.

igvm/hypervisor.py
kofrezo: I agree with @TuxPowered42 here. 

We have certain configuration requirements to a environment such as that some software is installed. This is part of the configuration management and should not be checked here. 

Adding such checks can be useful when having a transition from one OS to another (Stretch to Buster) in combination with conditionals but there is no need for this here.

igvm/hypervisor.py
```diff
@@ -826,8 +880,50 @@ def device_to_netcat(self, device, size, listener):
...
+        # xfsrestore needs to output its logs, otherwise it fails
+        self.run(
+            'nohup /bin/nc.openbsd -l -p {0} '
+            '| xfsrestore -F -J - {1} 2>{1}.log 1>&2 &'
+            .format(port, mount_dir),
+            pty=False,  # Has to be here for background processes
+        )
+        try:
+            yield self.fqdn, port
+        except BaseException:
```
kofrezo, line 908: Why are you explicelty capturing BaseException here ?

tests/test_integration.py
```diff
@@ -582,6 +582,45 @@ def test_offline_migration_drbd(self):
...
+        vm_migrate(
+            VM_HOSTNAME,
+            offline=True,
+            offline_transport='xfs',
+        )
+        self.check_vm_present()
+
+        disk_size_gib = self.vm_obj['disk_size_gib']
+
+        # Increase disk size during migration
```
kofrezo, line 596: This should be a dedicated test.

tests/test_integration.py
```diff
@@ -582,6 +582,45 @@ def test_offline_migration_drbd(self):
...
+        vm_migrate(
+            VM_HOSTNAME,
+            offline=True,
+            offline_transport='xfs',
+            disk_size=disk_size_gib + 1,
+        )
+        self.check_vm_present()
+
+        # Decreasing disk size back during migration
```
kofrezo, line 605: This should be a dedicated test.

tests/test_integration.py
```diff
@@ -582,6 +582,45 @@ def test_offline_migration_drbd(self):
...
+        vm_migrate(
+            VM_HOSTNAME,
+            offline=True,
+            offline_transport='xfs',
+            disk_size=disk_size_gib,
+        )
+        self.check_vm_present()
+
+        # Attempt to decrease disk size lower than allocated space
+        with self.assertRaises(IGVMError):
```
kofrezo, line 615: This should be a dedicated test.

tests/test_integration.py
```diff
@@ -643,6 +682,21 @@ def test_rollback_drbd(self):
 
         self.check_vm_present()
 
+    def test_rollback_xfs(self):
+        obj = Query({'hostname': VM_HOSTNAME}, ['puppet_environment']).get()
```
kofrezo, line 686: You should be able to access it via `self.vm_obj` without doing an extra Query.

igvm/hypervisor.py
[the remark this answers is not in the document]
Felixoid: More verbose description http://llewellynfalco.blogspot.com/2016/02/dont-use-greater-than-sign-in.html

kofrezo (changes requested):
Thanks for adding this cool new feature, I am looking forward to use it in production. 

There are some things missing such as doc strings and some changes I disagree on. Please check them and lets discuss.

igvm/cli.py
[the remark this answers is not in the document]
Felixoid: It is an optional argument, that works only together with `--offline --offline-transport xfs`.

If it's used without it:

```
  File "run.py", line 12, in <module>
    main()
  File "/home/mikhail.shiryaev/igvm/igvm/cli.py", line 440, in main
    args.pop('func')(**args)
  File "/home/mikhail.shiryaev/igvm/igvm/host.py", line 25, in decorator
    return fn(*args, **kwargs)
  File "/home/mikhail.shiryaev/igvm/igvm/commands.py", line 398, in vm_migrate
    _vm, offline, offline_transport, disk_size
  File "/home/mikhail.shiryaev/igvm/igvm/hypervisor.py", line 584, in migrate_disk_size
    self.check_migrate_vm(vm, offline, offline_transport, disk_size)
  File "/home/mikhail.shiryaev/igvm/igvm/hypervisor.py", line 571, in check_migrate_vm
    'disk_size can be applied only with offline transport xfs!'
igvm.exceptions.StorageError: disk_size can be applied only with offline transport xfs!
```

igvm/hypervisor.py
Felixoid: This method is an unchanged snippet from [1](https://github.com/innogames/igvm/pull/176/commits/5fff38a34b71643951dc72192c65625535f06240#diff-d587443efd90ee0cdc812d478981451c4b4a15000f584b1fc06ea3256d154fddL609-L619) and [2](https://github.com/innogames/igvm/pull/176/commits/5fff38a34b71643951dc72192c65625535f06240#diff-d587443efd90ee0cdc812d478981451c4b4a15000f584b1fc06ea3256d154fddL622-L632)

igvm/hypervisor.py
Felixoid: Don't get it. What actually should be a parameter? All of the arguments of this method are parameters of `migrate_vm`

igvm/hypervisor.py
Felixoid: The parameter is the CLI argument, so `no_shutdown` is coming from the parser. I'll rename the method though.

igvm/hypervisor.py
Felixoid: This is the same usage as in `netcat_to_device`. I suspect this is done to keep flake8 silent about `cathing everything` error.

tests/test_integration.py
Felixoid: True, but it's done in the same way for every rollback test. Should I change it for everything else too?

igvm/cli.py
```diff
@@ -185,6 +185,14 @@ def parse_args():
...
+        type=int,
+        help='When --offline-transport=xfs is used, this parameter may specify'
```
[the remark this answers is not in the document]
TuxPowered42: It is an optional argument, once it is specified it does things always. Saying "may" means that it could still be ignored depending on other conditions.

igvm/hypervisor.py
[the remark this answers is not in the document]
TuxPowered42: Such notation makes perfect sense for testing ranges. But for single numbers…? I dunno. @kofrezo is thre a PEP for that? I've checked PEP-8 and it's not specified there, however all examples do `VAR > VAL`, not the opposite.

igvm/cli.py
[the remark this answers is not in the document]
TuxPowered42: `0` is a valid (not for this particular option but I'm talking coding in general) value which user might enter. `None` means the option was not given at all. Those are separate things and using Magic Numbers is not necessary in languages like Python, where we can easily check if things exist at all. 

Back to the other piece of code, where it is used:
```
disk_size = disk_size if 0 < disk_size else 0
if not (offline and offline_transport == 'xfs') and 0 < disk_size:
```

Could just be:
```
if disk_size and disk_size < 1:
    raise StorageError('New disk size must be at least 1GiB!')
if disk_size and not (offline and offline_transport == 'xfs'):
    raise StorageError('Disk can only be resized with offline migration and --offline_transport=xfs!')
```

tests/test_integration.py
TuxPowered42: If every other test queries a new object, maybe it is necessary for something? I vaguely remember there was some magic happening here.

igvm/hypervisor.py
TuxPowered42: This function was designed for migration of Graphite hosts and they needed a very long time to shut down. Let's just keep it as it is.

tests/test_integration.py
Felixoid: Done

tests/test_integration.py
Felixoid: I would leave it as one test for disk changing since first it increases the disk and then decreases to the original size.

tests/test_integration.py
Felixoid: done

tests/test_integration.py
Felixoid: I'd like to make it consistent with other tests for now. If you want, I could leave a TODO comment for future

igvm/hypervisor.py
Felixoid: Changed. Looks better?

igvm/hypervisor.py
Felixoid: Renamed to `wait_for_shutdown`

igvm/hypervisor.py
Felixoid: Renamed to `vm_new_disk_size`

Felixoid:
I will squash commits tomorrow to ease the review since now it doesn't look pretty. So only 5 commits will be in place

igvm/hypervisor.py
kofrezo: Why not `apply_new_vm_disk_size` - this is more close to our every day English than `apply_vm_new_disk_size` ?

igvm/hypervisor.py
kofrezo: My bad I missed how it is used in migrate_vm. This is fine.

You could still add a doc string and consider making it a underscore method.

igvm/hypervisor.py
kofrezo: We should not allow changing the disk size without offline in the first place then. Similar to this: 
-  https://github.com/innogames/igvm/blob/master/igvm/commands.py#L418

Making the user think he can migrate the VM and change the disk size online and then ask him to shutdown the VM is counter intuitive and does not fit our already existing handling of similar cases. We are better of telling him right away that this is not possible.

igvm/hypervisor.py
kofrezo: Okay but lets at least tell the user than how long we will wait.

igvm/hypervisor.py
kofrezo: Let me rephrase the questions: Why do you expect yielding two primitive data types (a string and a integer) to raise a BaseException ?

If it is because of the method `_netcat_port` raising it we should consider using Transaction.

tests/test_integration.py
kofrezo: This are two scenarios for the user. One might increase a disk size with IGVM but never decrease it and vice versa.

Additionally having many small test methods allows us to run tests in parallel and will speed up the test pipeline.

tests/test_integration.py
kofrezo: Fair enough. Lets just make a todo.

kofrezo (changes requested):

igvm/hypervisor.py
Felixoid: I wanted to make it consistent with `vm_new_disk_size` -> `apply_vm_new_disk_size`. First vm indicates the scope.

Maybe, `_vm_apply_new_disk_size` would sound better. The method looks private to me.

igvm/hypervisor.py
Felixoid: From my side, I've just reused the same snippet from https://github.com/innogames/igvm/commit/7ed936d4#diff-d587443efd90ee0cdc812d478981451c4b4a15000f584b1fc06ea3256d154fddR743

Maybe, the contextmanager API is used wrong in the first place, but as far as I know, @TuxPowered42 would like to get rid of Transaction and to move to the context lib.

igvm/hypervisor.py
Felixoid: I guess the last comment is related to some other discussion.

Nevertheless, it looks more reasonable than to just replace https://github.com/innogames/igvm/blob/master/igvm/commands.py#L418 with `check_migrate_parameters`

tests/test_integration.py
Felixoid: I see the point. But the parallel execution has the overhead of VM deployment, and to be sure that the decrease operation will work, I have to increase the disk in advance.

If you insist, it's possible to apply `vm_set_disk_size_gib` before xfs migration with disk size decrease

igvm/hypervisor.py
Felixoid: Here's the answer, why `BaseException` is caught https://docs.python.org/3.5/library/contextlib.html#contextlib.contextmanager

> If an unhandled exception occurs in the block, it is reraised inside the generator at the point where the yield occurred.

So, if anything happens at https://github.com/innogames/igvm/pull/176/files#diff-d587443efd90ee0cdc812d478981451c4b4a15000f584b1fc06ea3256d154fddR694, the `@contextmanager` will clean up the netcat server on the target hypervisor.

igvm/hypervisor.py
Felixoid: And `check_migrate_parameters` are already included here https://github.com/innogames/igvm/pull/176/commits/0371e62cd780a9044b36498ac7971049410228fb#diff-9299a928e2b4f7d2b1d51a5dcdd44a442d587f35480a7f2dfe8349022180e2cdR397, so the migration won't start on the wrong parameters

igvm/hypervisor.py
Felixoid: Added to argparse help

tests/test_integration.py
Felixoid: Split the test into two different

tests/test_integration.py
Felixoid: TODO added as a dedicated commit

igvm/hypervisor.py
kofrezo: Yes I think that is much better.

igvm/hypervisor.py
kofrezo: Fair enough.

igvm/hypervisor.py
kofrezo: Ah makes sense. Thanks for explaining.

kofrezo (approved):

kofrezo:
We still need to figure out a way to make the newly added tests pass always because the disk resize sometimes fails (for a unknown reason) but for the status quo of code you have my green light.

kofrezo:
> We still need to figure out a way to make the newly added tests pass always because the disk resize sometimes fails (for a unknown reason) but for the status quo of code you have my green light.

Already tackled by:

- https://github.com/innogames/igvm/pull/176/commits/21c311efed2158b4b99241e3b85dac8ace016c92
- https://github.com/innogames/igvm/pull/176/commits/116db75c36702e430fcfe058a69b37f96f14516c

igvm/hypervisor.py
```diff
@@ -800,34 +885,143 @@ def check_netcat(self, port):
...
+    def _xfsrestore_log_name(self, vm: VM) -> str:
+        device_name = self.get_volume_by_vm(vm).name()
+        return '/tmp/xfsrestore-{}.log'.format(device_name)
+
+    def check_xfsrestore_log(self, vm: VM):
+        """
+        Search for WARNING in the xfsrestore log file.
+        Raises exception if found
+        """
+        self.run('cat {}'.format(self._xfsrestore_log_name(vm)))
```
TuxPowered42, line 954: Why first we cat the file and then grep it?

Felixoid: To have the log output in the migration log. Only for this.

TuxPowered42: There is no migration log. If you mean dumping the output to  console, that's fine.

TuxPowered42 (approved):

Felixoid:
Thank you for your review, gentlemen.

Felixoid:
@ig-sysadmin-jenkins retest this please

Felixoid:
@ig-sysadmin-jenkins retest this please

PR #186, opened 2020-12-07
Prefer shared credentials file for managing AWS

lacunoc:
Other tools, like e.g. restic,  might also use AWS_ exported
env variables, which will then break the administration
of AWS via igvm. So we need to prefer the shared credentials
file for all actions.


lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

kofrezo (commented):
Codewise it looks fine to me. I was just confused by the title of the PR. What is this default profile and why do we prefer it ?

igvm/vm.py
```diff
@@ -241,7 +242,8 @@ def aws_start(self):
         Start a VM in AWS.
         """
 
-        ec2 = boto3.client('ec2')
+        session = boto3.Session(profile_name=AWS_PROFILE_NAME)
```
kofrezo, line 245: Does it make sense to create the session once and store it as instance variable and use a getter to create it the first time and otherwise return it to avoid creating a new session every time in a run ?

kofrezo (approved):

lacunoc:
The default profile is the profile you can add to your users credentials file. It needs to be preferred, because otherwise exported ENV variables are used, which are wrong, because they are exported by restic on a global basis on every host.

igvm/vm.py
lacunoc: You're absolutely right. I reworked it using a property for the session initialisation.

lacunoc:
retest this please

konstantinbest (approved):

PR #188, opened 2020-12-08
Fix SSH connect errors during cloud_init provision

lacunoc:
We need to reduce the amount of our SSH connect tries, otherwise we a SSHException is thrown, which is not catchable.


lacunoc:
retest this please

konstantinbest (approved):

PR #189, opened 2021-01-05
Support multi region AWS setup

lacunoc:
We spreaded our network setup over 3 regions
and therefore we need to check, for which
region we need to execute the AWS API call.


konstantinbest (commented):
lgtm

konstantinbest (approved):

TuxPowered42 (approved):

PR #190, opened 2021-01-15
Add option to specify destination HV for evacuate

Woellchen:
This allows specifying a destination hypervisor in the evacuate
command. This is especially useful when one wants to move all VMs of
one hypervisor to another new, empty one for maintenance work.


lamaral (approved):
LGTM

PR #192, opened 2021-01-19
Add a nice progress bar for offline migrations

Woellchen:
This removes the deprecated usage of /proc/drbd to get the current
migration status and instead uses the "drbdsetup events2" command.
The new progress bar uses tqdm and will not print a new line each time
and will look approximately like this:
HOSTNAME:  70%|███   | 16.2G/23.0G [01:26<00:46, 158MiB/s]

### DIFF
--- a/igvm/drbd.py
+++ b/igvm/drbd.py
@@ -7,6 +7,10 @@ from contextlib import contextmanager
 from io import BytesIO
 from logging import getLogger
 from time import sleep
+from typing import Tuple, Optional
+
+from tqdm import tqdm
+
 from igvm.exceptions import RemoteCommandError
 
 log = getLogger(__name__)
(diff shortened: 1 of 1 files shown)


igvm/drbd.py
```diff
@@ -7,6 +7,10 @@
 from io import BytesIO
 from logging import getLogger
 from time import sleep
+from typing import Tuple, Optional
+
+from tqdm import tqdm
```
TuxPowered42, line 12: Will it handle non-terminal runs, like redirections to a file?

igvm/drbd.py
```diff
@@ -267,44 +271,57 @@ def replicate_from_master(self, transaction=None):
...
+        Optional[str],
+        Optional[int],
+        Optional[int],
+    ]:
+        """Returns a tuple containing statistics about the DRBD status."""
+        rep = None
+        size = None
+        oos = None
+
+        events = self.hv.run('drbdsetup events2 --now --statistics {}'.format(
```
TuxPowered42, line 305: Does it work on Stretch HVs too?

igvm/drbd.py
```diff
@@ -267,44 +271,57 @@ def replicate_from_master(self, transaction=None):
...
+        last_n_synced = 0
+
+        with tqdm(
+            desc=self.vm_name,
+            total=size,
+            unit='iB',
+            unit_scale=True,
+            unit_divisor=1024,
+        ) as pbar:
+            while rep != 'Established' and rep is not None:
```
TuxPowered42, line 284: The original code was handling a case where progress bar was not yet available because the receiving side has not yet started syncing. Is this handled with the new code too?

igvm/drbd.py
Woellchen: yes, tested a migration from stretch to buster and both hvs were showing the same output

igvm/drbd.py
Woellchen: we don't need to handle it anymore. as soon as you run "drbdadm up" the command will have statistics available

igvm/drbd.py
Woellchen: well yeah. it will print `\r` in the file though as it would on the terminal. is this a problem for you? will look ugly in `less` for example but print beautifully on your terminal with `cat` or `tail`

TuxPowered42: > 
> disable : bool, optional
>
>   Whether to disable the entire progressbar wrapper [default: False]. If set to None, disable on non-TTY.

Let's check setting it to `None`.

Woellchen: works, set to `None`

TuxPowered42 (approved):

lamaral (approved):

PR #193, opened 2021-01-20
Use Haswell-noTSX to remove missing features

lamaral:
Intel has disabled TSX on Haswell CPUs due to security vulnerabilities.
hle and rtm are two subfeatures of TSX and because of that, when trying
to spawn VMs on Hypervisors with a newer BIOS version, an error is
thrown because hle and rtm are not available.

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c2955f270a84762343000f103e0640d29c7a96f3


Woellchen (approved):

TuxPowered42 (approved):

lamaral:
restest this please

PR #195, opened 2021-02-02
Fix VM memory overhead calculation

Woellchen:
Previously, the VM memory overhead was only applied once for the
whole Hypervisor. Which is wrong, because every VM has its very own
memory overhead in QEMU.
This commit fixes this calculation and therefore prevents unintended
overallocation and/or swapping of VM memory on the Hypervisor.


TuxPowered42 (approved):

Woellchen:
retest this please

konstantinbest (approved):

PR #196, opened 2021-02-02
Add a flag to overrule Hypervisor preferences

Woellchen:
This adds a flag "--soft-preferences" to the build, migrate and
evacuate commands of igvm. Setting them causes the Hypervisor
selection algorithm to behave slightly different. Instead of
excluding Hypervisors if any of the preferences fails it will instead
keep them with an accordingly lower ranked overall score. This way we
can force all Hypervisors to be considered as targets while still
leaving the hard checks for later on to check if VMs really fit.
In emergency cases this can help emptying Hypervisors or just moving
a specific VM to the "less worst" available location. It can also be
used to force a VM build and squeeze it in somewhere.

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -10,6 +10,7 @@ from contextlib import contextmanager, ExitStack
 from ipaddress import ip_address
 from os import environ
 from time import sleep
+from typing import List, Optional
 
 from adminapi.dataset import Query
 from adminapi.filters import Any, StartsWith, Contains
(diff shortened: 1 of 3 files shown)


Woellchen:
retest this please

Woellchen:
retest this please

Woellchen:
retest this please

igvm/commands.py
```diff
@@ -942,7 +964,12 @@ def _get_hypervisor(hostname, allow_reserved=False):
 
 @contextmanager
 def _get_best_hypervisor(
-        vm, hypervisor_states, offline=False, enforce_vm_env=False):
+    vm,
+    hypervisor_states,
+    offline=False,
+    enforce_vm_env=False,
+    soft_preferences=False,
```
kofrezo, line 971: Can we add type hints here too ?

kofrezo (approved):
LGTM

igvm/commands.py
Woellchen: done here and in one more place :)

PR #200, opened 2021-02-10
Sort out AWS instance_types with no ipv6 support

lacunoc:
When we are crawling the AWS instances overview
file, we need to sort out potential instance_types
which do not have any ipv6 support, as it will break
the VM build with a sleep loop.


TuxPowered42 (approved):

konstantinbest (approved):

lacunoc:
retest this please

PR #201, opened 2021-02-22
Add cleanup for orphaned igvm test VMs in AWS

lacunoc:


lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

lacunoc:
retest this please

tests/conftest.py
```diff
@@ -124,57 +125,100 @@ def clean_serveradmin(filters):
...
 
+
+def clean_aws(vm_hostname):
     try:
         obj = Query({'hostname': vm_hostname}, ['aws_instance_id']).get()
     except DatasetError:  # No object to clean up
         return
 
+    timeout = 120
+    ec2 = boto3.client('ec2')
```
Woellchen, line 157: why exactly was this moved above the if? those variables are not required for the if right?

tests/conftest.py
```diff
@@ -124,57 +125,100 @@ def clean_serveradmin(filters):
...
     try:
         ec2.stop_instances(
             InstanceIds=[obj['aws_instance_id']], DryRun=False
         )
+        for _ in range(timeout):
+            instance_status = _get_aws_instance_status(obj['aws_instance_id'])
+            if AWS_RETURN_CODES['stopped'] == instance_status:
+                break
+            sleep(1)
     except ClientError as e:
```
Woellchen, line 170: this `ClientError` is too broad, especially now that we moved another API call in there. if you want to check for not running or missing instances, please use `Client.InvalidInstanceID.NotFound` and `InvalidInstanceID` or other types of errors if applicable. see https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html
all other errors are unexpected and should not be `pass`ed so we can see them

tests/conftest.py
```diff
@@ -124,57 +125,100 @@ def clean_serveradmin(filters):
...
     for _ in range(timeout):
-        instance_status = _get_instance_status()
+        instance_status = _get_aws_instance_status(obj['aws_instance_id'])
         if AWS_RETURN_CODES['terminated'] == instance_status:
             break
         sleep(1)
 
 
+def clean_orphaned_aws_vms():
+    def _get_leftover_vms() -> List:
```
Woellchen, line 182: this is almost the same code as in `_get_aws_instance_status`. we can make a shared function that takes optional filter arguments (instance id, ..)

tests/conftest.py
```diff
@@ -124,57 +125,100 @@ def clean_serveradmin(filters):
...
+                            str(AWS_RETURN_CODES['shutting-down']),
+                            str(AWS_RETURN_CODES['stopping']),
+                            str(AWS_RETURN_CODES['stopped']),
+                        ]
+                    },
+                ]
+            )['Reservations']
+
+            instance_ids = [r['Instances'][0]['InstanceId'] for r in response]
+        except IndexError:
```
Woellchen, line 204: error handling should not be abused for control flow. please make proper key or len checks and get rid of the `try` block

tests/conftest.py
```diff
@@ -124,57 +125,100 @@ def clean_serveradmin(filters):
...
+        instance_status = _get_aws_instance_status(obj['aws_instance_id'])
         if AWS_RETURN_CODES['terminated'] == instance_status:
             break
         sleep(1)
 
 
+def clean_orphaned_aws_vms():
+    def _get_leftover_vms() -> List:
+        try:
+            response = ec2.describe_instances(
```
Woellchen, line 184: calling `ec2` out of nowhere is pretty dangerous because it makes hard assumptions about it's scope. `ec2` should be passed as a parameter

tests/conftest.py
```diff
@@ -124,57 +125,100 @@ def clean_serveradmin(filters):
...
+            return []
+
+        return instance_ids
+
+    timeout = 120
+    ec2 = boto3.client('ec2')
+    leftover_vm_ids = _get_leftover_vms()
+    if len(leftover_vm_ids) > 0:
+        ec2.terminate_instances(InstanceIds=leftover_vm_ids)
+        for _ in range(timeout):
```
Woellchen, line 214: i counted this kind of loop now 3 times duplicated in this diff. lets move this out into a shared function

tests/conftest.py
```diff
@@ -124,57 +125,100 @@ def clean_serveradmin(filters):
...
+            instance_ids = [r['Instances'][0]['InstanceId'] for r in response]
+        except IndexError:
+            return []
+
+        return instance_ids
+
+    timeout = 120
+    ec2 = boto3.client('ec2')
+    leftover_vm_ids = _get_leftover_vms()
+    if len(leftover_vm_ids) > 0:
```
Woellchen, line 212: prime example for early return :)

tests/test_integration.py
```diff
@@ -104,6 +104,8 @@ def setUp(self):
 
         # Cleanup all leftovers from previous tests or failures.
         clean_all(self.route_network, self.datacenter_type, VM_HOSTNAME)
+        if self.datacenter_type == 'aws.dct':
```
Woellchen, line 107: any reason this should not go into `clean_all` function?

Woellchen (changes requested):

tests/test_integration.py
lacunoc: clean_all gets called more than once afais, so it will remove testing VMs, which are still in use for testing.

Woellchen: ah now understand. i maybe have a better idea in general to solve this. the actual problem is that SA objects might be missing that hold the instance_id right? so instead of making a whole new code branch we could instead solve only this one edge case. in `clean_aws()` there is this `try` block which catches the `DatasetError`. i think what we might really want to do is instead of returning, try to find the right instance id through aws api. what do you think? then we do not have to work around this and can keep all existing logic int `clean_all()` and `clean_aws()`

tests/conftest.py
lacunoc: refactored the cleanup

tests/conftest.py
lacunoc: refactored the cleanup code

tests/conftest.py
lacunoc: `ec2` is now passed as parameter

tests/conftest.py
lacunoc: moved to a shared function called `_wait_for_state_reached`

tests/conftest.py
lacunoc: early returning now

tests/test_integration.py
lacunoc: good idea, let's see, how this goes.

tests/conftest.py
lacunoc: raising the exception if its not a `InvalidInstanceID`

tests/conftest.py
lacunoc: refactored

lacunoc:
retest this please

lacunoc:
retest this please

konstantinbest (approved):

Woellchen (approved):

PR #202, opened 2021-02-24
Strip hostname out of FQDN for AWS user_data

lacunoc:
We need the FQDN and the hostname for initial cloud_init
setup, to put these into /etc/hosts after first boot.
The previous rstrip broke the hostname for specific projects.


konstantinbest (approved):
lgtm

PR #204, opened 2021-03-23
Automatically assign SGs to VMs in AWS

TuxPowered42:


TuxPowered42:
retest this please

igvm/vm.py
```diff
@@ -558,25 +622,25 @@ def aws_build(self,
...
         root_device = list(
-            ec2_res.images.filter(
+            self.ec2r.images.filter(
                 ImageIds=[self.dataset_obj['aws_image_id']]
             )
         )[0].root_device_name
         disk_size_gib = self.dataset_obj['disk_size_gib']
 
+        for aws_sg in self.aws_sgs:
+            self.dataset_obj['aws_security_group_ids'].add(aws_sg.group_id)
```
TuxPowered42, line 639: @lacunoc Should we rather replace manually assigned groups instead of appending to them?

lacunoc: Yes, we should.

tests/test_integration.py
```diff
@@ -132,6 +132,7 @@ def setUp(self):
             self.vm_obj['aws_key_name'] = 'eu-central-1-key'
             self.vm_obj['service_groups'] = [
                 'igvm-integration.test.sg',
+                'ig-aws-base-internet.sg',
```
lacunoc, line 135: We maybe should add a TODO here to get rid of this group.

lacunoc (changes requested):
beside of the minor comments, lgtm

igvm/vm.py
TuxPowered42: The attribute is now still populated and commited but overwritten. Removal of attributes will be in another MR.

tests/test_integration.py
TuxPowered42: Done.

lacunoc (approved):

PR #206, opened 2021-04-26
Use single AWS SG per VM

TuxPowered42:


TuxPowered42:
Repleaced with a version which uses Serveradmin Firewall API to configure SGs in AWS.

PR #208, opened 2021-05-07
tests: Move SGs for AWS to network objects

lacunoc:


lacunoc:
retest this please

tests/test_integration.py
```diff
@@ -130,12 +130,6 @@ def setUp(self):
             self.vm_obj['aws_image_id'] = 'ami-0e2b90ca04cae8da5'  # buster
             self.vm_obj['aws_instance_type'] = 't2.micro'
             self.vm_obj['aws_key_name'] = 'eu-central-1-key'
-            self.vm_obj['service_groups'] = [
-                'ig-aws-base-internet.sg', # TODO: remove once we have our own image
-                'ig-aws-base-networking.ndco.sg',
-                'igvm-integration.test.sg',
-                'puppet-aws.admin.sg',
-            ]
```
TuxPowered42, line 138: Do we really want it? We don't need to grant the whole testing network permissions to igvm-integration.

lacunoc: Actually, yes. And it was like this already. It's only added to the specific one, which is used `test-euc1-az1-int`

lacunoc:
retest this please

tests/test_integration.py
TuxPowered42: Now I see, we are not adding SGs for tests in `kvm.dct`, why should we in `aws.dct`.

TuxPowered42 (approved):

PR #219, opened 2022-01-20
NDCO-2518: make /tmp a tmpfs

TuxPowered42:


igvm/templates/etc/fstab
```diff
@@ -1,2 +1,3 @@
 proc /proc proc defaults 0 0
 /dev/{{ blk_dev }} / {{ type }} {{ mount_options }} 0 1
+tmpfs /tmp     tmpfs rw,nosuid,noexec,mode=01777,size=1G 0 0
```
Woellchen, line 3: do we need `nodev`?

TuxPowered42:
Systemd's `PrivateTmp` should provide adequate, if not better, security.

PR #228, opened 2022-05-09
Adjust Bullseye online migration support

lamaral:


igvm/settings.py
```diff
@@ -84,12 +90,10 @@
...
-    # Bullseye migrations. Only the straightforward ones that are likely
-    # to work for now. They are yet to be verified. Bullseye to buster must
-    # be figured out later.
-    # TODO: rememberme
-    ('bullseye', 'bullseye'): P2P_MIGRATION,
-    ('buster', 'bullseye'): P2P_MIGRATION,
+    # Online migrations are only working between Bullseye and Bullseye.
+    # The other cases for migrations (Buster to Bullseye and vice-versa) are
+    # not working, due to what looks like libvirt incompatibility.
+    ('bullseye', 'bullseye'): P2P_TLS_MIGRATION,
```
kofrezo, line 96: You may want a single test case for Debian Bullseye migration or maybe already switch all of them already [here](https://github.com/innogames/igvm/blob/master/tests/test_integration.py#L119).

kofrezo (approved):

igvm/settings.py
lamaral: The changes on this PR  are to enable migration between Bullseye hypervisors. 
The testing for migrations on Bullseye happens when we have the testing hypervisors on Bullseye, which is soonish.

kofrezo: okay

TuxPowered42 (approved):

PR #230, opened 2022-06-15
Unlock hypervisor early during build

Phantuman:
We will lock the hypervisor only for the critical
parts of the vm creation. This will allow more builds
to run in parallel.

### DIFF
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -599,6 +599,18 @@ class VM(Host):
                 )
                 self.hypervisor.download_and_extract_image(image, mount_path)
 
+            # This needs to happen at this point, so that the memory
+            # usage of this VM is already taken into consideration if
+            # a hypervisor is selected during a later igvm build run.
+            hypervisor.define_vm(self, transaction)
+
+            # Unlocking here is safe, because we keep the state in the
+            # hypervisor object and a second release will trigger no
+            # update on serveradmin.
+            self.hypervisor.release_lock()
+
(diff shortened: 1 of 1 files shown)


igvm/vm.py
```diff
@@ -599,6 +601,8 @@ def build(
                 )
                 self.hypervisor.download_and_extract_image(image, mount_path)
 
+                self.hypervisor.release_lock()
```
lamaral, line 604: The hypervisor will never be unlocked if somebody is building a barebones VM.

Phantuman: This function is executed inside a context manager, which will unlock the hypervisor always the latest that the function finshes

igvm/vm.py
```diff
@@ -599,6 +599,15 @@ def build(
                 )
                 self.hypervisor.download_and_extract_image(image, mount_path)
 
+            # This needs to happen at this point, so that the memory
+            # usage of this VM is already taken into consideration if
+            # a hypervisor is selected during a later igvm build run.
+            hypervisor.define_vm(self, transaction)
+
+            self.hypervisor.release_lock()
```
TuxPowered42, line 607: Let's add a comment stating why this double release of lock is safe.

TuxPowered42 (changes requested):

TuxPowered42 (approved):

Phantuman:
test this please

kofrezo:
test this please.

PR #231, opened 2022-07-06
Increase timeout for shutdown of vms in AWS

konstantinbest:
The shutdown of a AWS machine can take longer than 120 seconds during our disaster recovery.

### DIFF
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -383,7 +383,7 @@ class VM(Host):
                 force_stop_failed=check_vm_up_on_transaction,
             )
 
-    def aws_shutdown(self, timeout: int = 120) -> None:
+    def aws_shutdown(self, timeout: int = 240) -> None:
         """AWS shutdown
 
         Shutdown a VM in AWS.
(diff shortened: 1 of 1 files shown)


kofrezo (approved):

igvm/vm.py
```diff
@@ -383,7 +383,7 @@ def shutdown(self, check_vm_up_on_transaction=True, transaction=None):
                 force_stop_failed=check_vm_up_on_transaction,
             )
 
-    def aws_shutdown(self, timeout: int = 120) -> None:
+    def aws_shutdown(self, timeout: int = 240) -> None:
```
lacunoc, line 386: I'm not sure, if you want to rely on an _guessed_ increased timeout value in a DR case.
What about forcing a instance stop as a last resort instead of having a running instance left.
```
            response = self.ec2c.stop_instances(
                InstanceIds=[self.dataset_obj['aws_instance_id']],
                DryRun=False,
                Force=True
            )
```

Reference: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.stop_instances

lacunoc (changes requested):

PR #232, opened 2022-07-07
Add error handling for unsupported instance types

konstantinbest:
It may happen that our own calculation of the optimal instance type is not sufficient and therefore we do not find a suitable
instance type in the region. Therefore we have this failback mechanism, which uses a hardcoded instance type, which should
be available in most cases.

### DIFF
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -17,7 +17,7 @@ from hashlib import sha1, sha256
 from io import BytesIO
 from pathlib import Path
 from re import compile as re_compile
-from typing import Optional, List, Union
+from typing import Optional, List, Union, Tuple
 from uuid import uuid4
 
 import boto3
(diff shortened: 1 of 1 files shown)


igvm/vm.py
```diff
@@ -672,54 +740,22 @@ def aws_build(self,
         disk_size_gib = self.dataset_obj['disk_size_gib']
 
         for vm_type in vm_types:
```
lacunoc, line 742: What about just appending `c5.xlarge` to the `vm_types` list? Would be less code with the same result. I would also load the vm_type from the settings and make this a variable, as there's already `c6i` (Intel) and `c6a` (AMD) available, so you have a central place to replace the failback vm type.

lacunoc (approved):

konstantinbest:
We solved this issue already with a different PR: https://github.com/innogames/igvm/blob/master/igvm/vm.py#L665

PR #235, opened 2022-08-17
Fix xfs offline transfer for bullseye

Woellchen:
Some software in bullseye got a bit more picky with handling default
file handles, like stderr. I suppose the behavior is actually correct
now on bullseye.
Previously the command would return just fine and detach to the
background and the ssh channel would close. However, we never
redirected stderr of the nc.openbsd command, so it was still attached
to the ssh channel. On bullseye this will now keep the channel open.
This leads to the command never returning and blocking the execution
flow forever, resulting in a stuck migration.
The fix is to add a stderr redirection to the nc.openbsd command.

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -970,8 +970,8 @@ class Hypervisor(Host):
         #  -J: inhibits inventory update
         # xfsrestore needs to output its logs, otherwise it fails
         self.run(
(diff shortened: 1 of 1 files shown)


lamaral (approved):

Woellchen:
test this please

Woellchen:
test this please

PR #236, opened 2022-08-25
Remove numa_mode distinction

lamaral:
We have no use cases for numa_mode auto.

The code for pinning vCPUs in spread mode will generate the same
configuration as unbound mode when ran against hypervisors with a single
NUMA node.

Given all that, it makes no sense to differentiate numa_nodes anymore,
especially given that libvirt itself makes no distinction.

By consequence, we also fix an issue where unbound nodes couldn't be
migrated from a hypervisor with higher CPU count to a hypervisor with
lower CPU count, due to re-pinning being skipped.


Woellchen (approved):
LGTM

PR #237, opened 2022-09-30
Use limit function for possible faulty values on HVs

seqizz:
It would make to skip a hypervisor with a faulty value instead of crashing the whole selection process because of it. Normalization on the next step on `HypervisorAttributeValue` and `HypervisorAttributeValueLimit` functions currently naively assume that the values are below 100.
Relates to: GOP-27242


Woellchen:
restest this please

Woellchen (approved):

PR #238, opened 2022-10-05
Fix AWS overview file permission issues

konstantinbest:
We had problems with the permissions of the AWS overview files in the /tmp folder when the etag file changed and different users were working with AWS at the same time. This change creates the overview files for each user in theire homefolder. This way we don't run the risk of having permission problems.

### DIFF
--- a/igvm/settings.py
+++ b/igvm/settings.py
@@ -206,10 +206,9 @@ AWS_RETURN_CODES = {
     'stopped': 80,
 }
 
-AWS_GRP_NAME = environ.get('AWS_GRP_NAME', default='adm')
 AWS_INSTANCES_OVERVIEW_URL = 'https://www.ec2instances.info/instances.json'
-AWS_INSTANCES_OVERVIEW_FILE = '/tmp/AWS_INSTANCES_OVERVIEW_FILE.json'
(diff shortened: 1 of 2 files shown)


v0devil (approved):

igvm/settings.py
```diff
@@ -206,10 +206,9 @@
     'stopped': 80,
 }
 
-AWS_GRP_NAME = environ.get('AWS_GRP_NAME', default='adm')
 AWS_INSTANCES_OVERVIEW_URL = 'https://www.ec2instances.info/instances.json'
-AWS_INSTANCES_OVERVIEW_FILE = '/tmp/AWS_INSTANCES_OVERVIEW_FILE.json'
-AWS_INSTANCES_OVERVIEW_FILE_ETAG = '/tmp/AWS_INSTANCES_OVERVIEW_FILE.etag'
+AWS_INSTANCES_OVERVIEW_FILE = 'AWS_INSTANCES_OVERVIEW_FILE.json'
+AWS_INSTANCES_OVERVIEW_FILE_ETAG = 'AWS_INSTANCES_OVERVIEW_FILE.etag'
```
TuxPowered42, line 211: This will still cause problems if the user wants to build multiple VMs in parallel. Please use https://docs.python.org/3/library/tempfile.html instead.

PR #239, opened 2022-10-10
Fix disk_size checks with xfs offline transfer

Woellchen:
When the VMs disk size was too big for the target Hypervisor, even though the requested disk size during a xfsdump migration was small enough, the sanity checks would fail. This was because the internal VMs state was reset too early.
Moving the disk size reset to a later stage fixes the issue.


lamaral (approved):
LGTM

PR #241, opened 2023-01-10
Improve message waiting for memory to become visible

kofrezo:
We exit when trying to shrink memory on a online VM a few lines before and we tell the user that we cannot rollback the memory if it failed a few lines afterwards in the sanity check so there is no point of spaming the user here (anymore).


kofrezo:
No test needed for this one. I just adjusted a log message text.

ragged (approved):

PR #243, opened 2023-01-27
Fix hypervisor attribute limit warning

konstantinbest:
This message was originally intended to sort out a hypervisor if the limit for an attribute is exceeded (for example CPU usage higher than 100). 

However, the log message was incorrect (string format passed 3 attributes, but 4 were expected.) and caused igvm to abort with an "IndexError: tuple index out of range".


seqizz (approved):
fml

PR #244, opened 2023-02-01
Ensure selection of fallback instance type if necessary

konstantinbest:
It may happen that we cannot find an instance type for a VM because the load values are too low. In this case, the build would simply stop. This change ensures that we always have at least the fallback instance type when installing.

### DIFF
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -659,9 +659,9 @@ class VM(Host):
         """
 
         vm_types_overview = self.aws_get_instances_overview()
-        if vm_types_overview:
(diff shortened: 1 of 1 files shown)


igvm/vm.py
```diff
@@ -659,9 +659,9 @@ def aws_build(self,
         """
 
         vm_types_overview = self.aws_get_instances_overview()
-        if vm_types_overview:
```
v0devil, line 662: Seems aws_get_instances_overview can return None, and then there is iteration aws_get_fitting_vm_types through overview var which can be None and then probably lead to TypeError.

v0devil (changes requested):

PR #246, opened 2023-02-17
Add retry logic to failed libvirt connections

Woellchen:


kofrezo (approved):

kofrezo:
Looks good to me. Have you tested it already in the wild ?

Woellchen:
Yes I have tested a couple of migrations where I randomly killed libvirt connections to either one or both HVs and they were able to recover and finish successfully. Same with rollbacks where I randomly killed connections before aborting the migrations and the rollbacks finished successfully as well.

PR #253, opened 2023-06-26
Adjust hypervisor state check logic

lamaral:
As we now have the cold_standby state, it doesn't make sense to check for retired only.

This was an issue for hypervisors in maintenance state as well, as igvm should never operate on them.

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -57,7 +57,7 @@ class Hypervisor(Host):
     def __init__(self, dataset_obj):
         super(Hypervisor, self).__init__(dataset_obj)
 
-        if dataset_obj['state'] == 'retired':
+        if dataset_obj['state'] not in ['online', 'online_reserved']:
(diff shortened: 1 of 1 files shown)


igvm/hypervisor.py
```diff
@@ -57,7 +57,7 @@ class Hypervisor(Host):
     def __init__(self, dataset_obj):
         super(Hypervisor, self).__init__(dataset_obj)
 
-        if dataset_obj['state'] == 'retired':
+        if dataset_obj['state'] not in ['online', 'online_reserved']:
```
TuxPowered42, line 60: The maintenance and deploy states must still allow to create the object. This is needed to be able to connect to a hypervisor to move VMs out of it.

lamaral: In `deploy_online` nothing should be happening on the hypervisor, but for maintenance it makes sense.

PR #258, opened 2023-10-10
Validate relevant Serveradmin attributes of VMs in AWS before build

konstantinbest:


igvm/vm.py
```diff
@@ -202,7 +202,8 @@ def revert_hostname(self):
     def check_serveradmin_config(self):
         """Validate relevant Serveradmin attributes"""
 
-        mul_numa_nodes = 128 * self.hypervisor.num_numa_nodes()
+        mul_numa_nodes = 128 * self.hypervisor.num_numa_nodes() \
+            if self.hypervisor else 1
```
TuxPowered42, line 206: The `ś=ż if ń else ź` format is not really readable for lines with breaks. Please use the normal `if … \n … else \n … \n`.

TuxPowered42 (approved):

PR #262, opened 2023-12-18
Skip instance_type selection if type is already set

konstantinbest:
The current solution for figuring out the best instance_type is not scalable for the disaster recovery case because we are parsing a 70 MB big json file in parallel for every VM. We are currently working on a different solution to prefill the instance_type for the disaster recovery case. We'll keep the functionality in igvm for now to be able to build VMs in AWS without the need of a prefill.

### DIFF
--- a/igvm/vm.py
+++ b/igvm/vm.py
@@ -28,7 +28,7 @@ from fabric.api import cd, get, hide, put, run, settings
 from fabric.contrib.files import upload_template
 from fabric.exceptions import NetworkError
 from json.decoder import JSONDecodeError
-from urllib.error import HTTPError
+from urllib.error import HTTPError, URLError
 from urllib.request import Request, urlopen
 
 from igvm.exceptions import ConfigError, HypervisorError, RemoteCommandError, VMError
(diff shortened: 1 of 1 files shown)


igvm/vm.py
```diff
@@ -713,12 +713,21 @@ def aws_build(self,
...
 
-        vm_types_overview = self.aws_get_instances_overview()
-        if vm_types_overview:
-            vm_types = self.aws_get_fitting_vm_types(vm_types_overview)
+        # The current solution for figuring out the best instance_type is not
+        # scalable for the disaster recovery case because we are parsing a
+        # 70 MB big json file in parallel for every VM. We are currently working
+        # on a different solution to prefill the instance_type for the disaster
+        # recovery case. We'll keep the functionality in igvm for now to be able
+        # to build VMs in AWS without the need of a prefill.
```
TuxPowered42, line 721: What is exactly the problem here? Is it loading the json file into Python structures or parsing those structures?

konstantinbest: The memory footprint of loading the file is too big.  Building 10 machines in parallel that way, the server allocates more then 2GiB of memory and we want to be able to build at least 50 VMs in parallel on one server.

PR #267, opened 2024-03-14
Force memory sizes to conform to system memory boundaries

TuxPowered42:


igvm/vm.py
```diff
@@ -233,6 +235,27 @@ def check_serveradmin_config(self):
             ('puppet_master', lambda v: True, 'puppet_master must be set'),
         ]
 
+        # Hosts defined with topmost address higher than MEM_BLOCK_BOUNDARY_GiB will use
+        # 1GiB or 2GiB memory block size. There is always extra 1GiB address space for
+        # the PCI bus. A dhost host defined with an even amount of memory ends up with
```
kofrezo, line 240:
```suggestion
        # the PCI bus. A host host defined with an even amount of memory ends up with
```

kofrezo (approved):

igvm/vm.py
TuxPowered42: Fixed.

PR #268, opened 2024-04-05
vcpu_set: Implement plus-minus operators

seqizz:
This was disturbing me for a while. Shouldn't be this easy though, please advise if I missed anything.


seqizz:
We have even spicier bug: If you try to add 2000 vcpus online, it doesn't fail out until it undefines the VM, leaving you with a broken state (and no vm). Will try to tackle that too.

PR #270, opened 2024-05-31
DRAFT: Enable AVX512 cpu features on guests if available on host

dreis2211:
Hi 👋 

This is largely untested, but a draft of what I believe would be needed to pass down CPU flags down to the guest if the host has them available. Mostly only relevant for Intel AVX 512 at this point, because we don't seem to have any Zen 4 Generation AMD hosts.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/igvm/kvm.py
+++ b/igvm/kvm.py
@@ -388,6 +388,7 @@ def generate_domain_xml(hypervisor, vm):
     if props.qemu_version >= (2, 3):
         _set_cpu_model(hypervisor, vm, tree)
         _place_numa(hypervisor, vm, tree, props)
+        _enable_cpu_features(tree)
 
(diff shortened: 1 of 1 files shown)


lamaral:
I will see if I can test it this week, as we got one Genoa server available.

dreis2211:
I think the most important thing to test is that VMs can be migrated between hosts that support these flags and those that don't, so that migration capabilities are not disturbed. That's what I would expect from the `optional` policy, but you never know.

In other words. In the best possible case we would win the additional support for new and shiny CPU flags for VMs on hosts that support it, and don't loose anything in all other situations.

lamaral:
While the optional policy allows you to start a domain on hypervisors missing these features, you still cannot live migrate VMs from a hypervisor with the features to a hypervisor without the features, as you can't remove features on the fly.

PR #275, opened 2024-08-13
Pin markupsafe dependency to 1.1

kofrezo:
Markupsafe 1.1.1 is the latest version shipped with Debian Bullseye which we are currently using.

Not specifying a version makes us run into a bug where markupsafe removed a dependency in a minor version (semver violation) when using jinja2 in version 2.x.

See: https://github.com/pallets/markupsafe/issues/304

This can be removed once we are on Debian Bookworm using Jinja 3.x

### DIFF
--- a/igvm/__init__.py
+++ b/igvm/__init__.py
@@ -1,6 +1,6 @@
 """igvm - Main Module
 
-Copyright (c) 2022 InnoGames GmbH
+Copyright (c) 2024 InnoGames GmbH
 """
 
-VERSION = (2, 2, 1)
+VERSION = (2, 2, 2)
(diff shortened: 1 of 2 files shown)


lamaral (approved):

kofrezo:
test this please.

kofrezo:
There is a single test failing in the test environment. But since we almost never use this command and it works on production (tested it manually) I am gonna skip this test here for now.

kofrezo:
Found out that the test breaks in the test environment under Debian Bookworm but works fine with Debian Bullseye. This is unrelated to this changes.

PR #276, opened 2024-08-14
Add rolling as release name

seqizz:
Small one which puts the rolling name as acceptable OS variant.

BOP-46882


kofrezo (approved):
I cannot judge on if this covers all necessary "corners" but as far as I can see this seems lgtm.

seqizz:
Thanks. I am sure more corner cases will haunt us later :sweat_smile:

PR #278, opened 2024-09-18
Support different VG names

seqizz:
Kind of related with NDCO-4920, this should allow us to build VMs on specific VGs.

Can be used to create mission-specific pools and assign VMs into it. As a first step it does not support migration, but that can be added later if need arise.

### DIFF
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -987,6 +987,14 @@ def _get_vm(hostname, unlock=True, allow_retired=False, vg_name=None):
                     hostname,
                 )
             )
+        if (
+            dataset_obj['libvirt_pool_override'] is not None
+            and dataset_obj['hypervisor'] is None
+        ):
+            raise InvalidStateError(
+                'Automatic HV selection is not possible for VMs with special '
+                'VG name requirements. Please specify a HV manually.'
+            )
         yield vm
     except (Exception, KeyboardInterrupt):
         VM(vm_query(), hypervisor).release_lock()
--- a/igvm/commands.py
+++ b/igvm/commands.py
@@ -37,6 +37,7 @@ from igvm.settings import (
     HYPERVISOR_ATTRIBUTES,
     HYPERVISOR_PREFERENCES,
     VM_ATTRIBUTES,
+    DEFAULT_VG_NAME,
 )
 from igvm.transaction import Transaction
 from igvm.utils import parse_size, parallel
@@ -319,6 +320,7 @@ def vm_build(
     soft_preferences: bool = False,
     barebones: bool = False,
     target_hv_query: Optional[str] = None,
+    vg_name: str = '',
 ):
     """Create a VM and start it
 
@@ -326,7 +328,10 @@ def vm_build(
     """
 
     with ExitStack() as es:
-        vm = es.enter_context(_get_vm(vm_hostname))
+        vm = es.enter_context(_get_vm(
+            hostname=vm_hostname,
+            vg_name=vg_name  # Need to pass only while building
+        ))
 
         if vm.dataset_obj['datacenter_type'] == 'aws.dct':
             # check if aws_image_id is our own, if yes, skip puppet_run in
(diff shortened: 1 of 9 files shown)


igvm/commands.py
```diff
@@ -399,6 +404,7 @@ def vm_build(
                 postboot=postboot,
                 cleanup_cert=rebuild,
                 barebones=barebones,
+                vg_name=vm.vg_name,
```
lamaral, line 407: You don't need to pass it here, as you can get it from the VM object itself.
```suggestion
```

igvm/hypervisor.py
```diff
@@ -65,17 +65,31 @@ def __init__(self, dataset_obj):
         self._mount_path = {}
         self._storage_type = None
 
-    def get_storage_pool(self) -> virStoragePool:
+    def get_active_storage_pools(self):
+        # Return all active storage pools
+        return self.conn().listAllStoragePools(2)
```
lamaral, line 70:
```suggestion
        # The 2 used as argument is the value of the VIR_CONNECT_LIST_STORAGE_POOLS_ACTIVE flag.
        return self.conn().listAllStoragePools(flags=2)
```

I think the comment change and specifying the keyword flags make it easier to understand.

igvm/hypervisor.py
```diff
@@ -701,19 +728,24 @@ def migrate_vm(
...
+                ) as listener:
                     self.xfsdump(vm, listener, transaction)
 
                 target_hypervisor.wait_for_xfsrestore(vm)
                 target_hypervisor.check_xfsrestore_log(vm)
                 target_hypervisor.umount_vm_storage(vm)
 
-            target_hypervisor.define_vm(vm, transaction)
+            target_hypervisor.define_vm(
+                vm=vm, transaction=transaction, vg_name=target_vg_name
```
lamaral, line 741:
```suggestion
                vm=vm, transaction=transaction
```

`define_vm()` has no vg_name argument. `target_vg_name` is also undefined.

igvm/hypervisor.py
```diff
@@ -701,19 +728,24 @@ def migrate_vm(
                     )
             elif offline_transport == 'xfs':
                 self._wait_for_shutdown(vm, no_shutdown, transaction)
-                with target_hypervisor.xfsrestore(vm, transaction) as listener:
+                with target_hypervisor.xfsrestore(
+                    vm=vm, transaction=transaction, vg_name=target_vg_name
```
lamaral, line 732:
```suggestion
                    vm=vm, transaction=transaction
```
xfsrestore() has no vg_name argument. target_vg_name is also undefined.

igvm/hypervisor.py
```diff
@@ -796,7 +828,7 @@ def redefine_vm(self, vm, new_fqdn=None):
         # XXX: undefine_vm depends on vm.fqdn beeing the old name for finding
         # legacy domains w/o an uid_name.  The order is therefore important.
         vm.fqdn = new_fqdn or vm.fqdn
-        self.define_vm(vm)
+        self.define_vm(vm=vm, vg_name=vm.vg_name)
```
lamaral, line 831:
```suggestion
        self.define_vm(vm)
```

`define_vm()` has no vg_name argument.
You can also fetch the value inside the function directly with `vm.vg_name`.

igvm/vm.py
```diff
@@ -79,13 +80,29 @@ def __init__(self, dataset_obj, hypervisor=None):
         # directly on running VM.
         self.mounted = False
 
+        if vg_name is None:
+            # Check if we can find it
+
+            if hypervisor is None:
+                # This is probably caused by an error,
+                # but we will pass as None
```
lamaral, line 88: This is actually the case for AWS, as we don't care about the hypervisors there. 
Please amend the comment, as it's not an error per se.

igvm/vm.py
```diff
@@ -646,6 +661,7 @@ def build(
             postboot=None,
             cleanup_cert=False,
             barebones=False,
+            vg_name=DEFAULT_VG_NAME,
```
lamaral, line 664:
```suggestion
```

You don't need the vg_name here, because you can use `self.vg_name` instead.

lamaral (approved):

PR #288, opened 2025-08-12
Add error handling for failing migration logging

konstantinbest:
It can happen, that multiple migrations at the same time will write the same log entry. This will fail during the commit and throw a DatasetError. This PR will add a retry mechanism.

### DIFF
--- a/igvm/hypervisor.py
+++ b/igvm/hypervisor.py
@@ -5,6 +5,9 @@ Copyright (c) 2018 InnoGames GmbH
 
 import logging
 import math
+
+from adminapi.dataset import DatasetError
+
 from contextlib import contextmanager
 from time import sleep, time
 from xml.etree import ElementTree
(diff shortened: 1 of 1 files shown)


igvm/hypervisor.py
```diff
@@ -1194,9 +1197,32 @@ def log_migration(self, vm: VM, operator: str) -> None:
...
-        self.dataset_obj.commit()
+        retry = 5
+        for i in range(0, retry):
+            # It can happen, that multiple migrations at the same time will write the same log
+            # entry. This will fail during the commit and throw a DatasetError
+            try:
+                cpu_usage_vm = self.estimate_vm_cpu_usage(vm)
+                timestamp = int(time())
+                log_entry = '{} {}{}'.format(timestamp, operator, round(cpu_usage_vm))
+                self.dataset_obj['igvm_migration_log'].add(log_entry)
```
kofrezo, line 1208: After this the changes are staged for commit so the first retry will add two estimated, the initial and the one from the retry, the second three and so on. I think you don't want this no ?

igvm/hypervisor.py
```diff
@@ -1194,9 +1197,32 @@ def log_migration(self, vm: VM, operator: str) -> None:
...
+                return
+            except DatasetError:
+                log.error(
+                    'Failed to log migration for VM "{}" on Hypervisor "{}". '
+                    'Will try again...'.format(
+                        str(vm),
+                        str(self),
+                    )
+                )
+                sleep(1)
```
kofrezo, line 1219: Just use the number of retry (i) as sleep duration this way the back of time in between increases which makes it a bit more graceful.

igvm/hypervisor.py
konstantinbest: I updated the code. In case the entry already exists, I'm removing it from the set again. The next iteration of the loop, will create a different entry, since the timestamp changed, which then will be unique again.

PR #296, opened 2026-02-12
Migrate old "Fabric3" -> "fabric" version 3.2 

brainexe:
Our current Fabric3 is quite outdated and causing more and more compatibility issues (like python3.13)
-> use the maintained "fabric" version

already tested: basic VM start/stop/build/delete/sync/info/restart/disk-set/mem-set
todo: migration (offline + online) / AWS / rename / vcpu-set

### DIFF
--- a/igvm/host.py
+++ b/igvm/host.py
@@ -37,7 +37,11 @@ class CommandResult(str):
     """
 
     def __new__(cls, result):
-        stdout = result.stdout if result.stdout else ''
+        # Fabric 1.x automatically stripped trailing whitespace from the
+        # string representation.  Fabric 3.x does not, so we strip here
+        # to avoid embedded newlines breaking commands that interpolate
+        # the result (e.g. mktemp output used as a mount path).
+        stdout = result.stdout.strip() if result.stdout else ''
         instance = super().__new__(cls, stdout)
         instance._result = result
         return instance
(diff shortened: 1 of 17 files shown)


igvm/host.py
```diff
@@ -61,69 +107,130 @@ def match_uid_name(self, uid_name):
...
+
+        # Fabric 3's sudo() prepends "sudo -S -p '...' " directly to the
+        # command without wrapping it in a shell, unlike Fabric.
+        # This breaks commands containing shell constructs (while, for, if,
+        # pipes, etc.).  Wrap in "bash -c '...'" to restore the old behavior.
+        if with_sudo and args:
+            cmd = args[0]
+            # Escape single quotes for bash -c '...' wrapping
+            cmd = cmd.replace("'", "'\"'\"'")
+            args = ("bash -c '{}'".format(cmd),) + args[1:]
```
TuxPowered42: Maybe better use `shlex.quote` instead of manually crafting string replacements? Or just never run shell scripts and do things directly in Python instead.

igvm/host.py
```diff
@@ -132,13 +239,13 @@ def put(self, remote_path, local_path, mode='0644'):
...
         """
-        with self.fabric_settings():
-            tempfile = '/tmp/' + str(uuid4())
-            fabric.api.put(local_path, tempfile)
-            self.run(
-                'mv {0} {1} ; chmod {2} {1}'
-                .format(tempfile, remote_path, mode)
-            )
+        conn = self._get_connection()
+        tempfile = '/tmp/' + str(uuid4())
```
TuxPowered42, line 220: There is mktemp in Python, maybe it could be used here.

igvm/vm.py
```diff
@@ -1115,23 +1125,23 @@ def aws_disk_set(self, size: int, timeout_disk_resize: int = 60) -> None:
...
-                    log.info(
-                        'successfully resized disk of {} to {}GB'.format(
-                            self.dataset_obj['hostname'], size)
-                    )
-                    return
-
-            raise VMError('disk resize for {} failed'.format(
-                self.dataset_obj['hostname'])
+        disk_resize = self.run(
+            'growpart /dev/{} 1'.format(disk), warn_only=True,
```
TuxPowered42: Let's use modern python f-strings.

igvm/host.py
brainexe: in this case we create the /tmp/ on the target host. with mktemp we can only to it locally. I keep it as it is to keep changes (kinda) low

lamaral (approved):
I did some limited testing and couldn't break it, so LGTM.

TuxPowered42 (approved):

PR #297, opened 2026-02-20
update VERSION to 2.5.0

brainexe:
the current version is still 2.2.4 and was not pumped since a while

`ERROR: Could not find a version that satisfies the requirement igvm<3,>=2.3 (from versions: 2.1, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5)`

### DIFF
--- a/igvm/__init__.py
+++ b/igvm/__init__.py
@@ -3,4 +3,4 @@
 Copyright (c) 2024 InnoGames GmbH
 """
 
-VERSION = (2, 2, 4)
+VERSION = (2, 5, 0)


igvm/__init__.py
```diff
@@ -3,4 +3,4 @@
 Copyright (c) 2024 InnoGames GmbH
 """
 
-VERSION = (2, 2, 4)
+VERSION = (2, 5, 0)
```
kofrezo, line 6: Why skipping 3 versions and not 2.3.0 ?

kofrezo (approved):

igvm/__init__.py
brainexe: to be in sync with the tags again, else it might be more confusion I'd guess: https://github.com/innogames/igvm/tags

brainexe:
I'll bumb the version in other active changes

PR #302, opened 2026-03-11
Calculate full XFS size, not usable size

TuxPowered42:
When building VMs we require a specific disk size and that ends up being the block device size. When a filesystem is created in this block device it might have some overhead. This overhead was quite small on older Debian releases but on Trixie it has grown quite a bit. The growth is mostly attributed to the log. It was ~160MiB before and is ~371MiB now.

Update the test function _get_disk_vm() to get the raw XFS size, because that's the one which should be identical to the block device size.

### DIFF
--- a/igvm/settings.py
+++ b/igvm/settings.py
@@ -70,12 +70,19 @@ KVM_HWMODEL_TO_CPUMODEL = {
     'EPYC': ['Dell_R6515', 'Dell_R7515', 'Supermicro_H13S'],
 }
 
+# Compatibility options used when making a new filesystems.
+# This is configured per VM's operating system because the VM's kernel
+# must be able to boot from and mount the FS. But we also limit it
+# to the oldest Hypervisor we have, because the HV must be able to
+# make the filesystem and mount it for building or for VM offline
+# XFS-aware migrations and disk shrinking.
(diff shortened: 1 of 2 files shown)


kofrezo (approved):

kofrezo:
> When building VMs we require a specific disk size and that ends up being the block device size. When a filesystem is created in this block device it might have some overhead. This overhead was quite small on older Debian releases but on Trixie it has grown quite a bit. The growth is mostly attributed to the log. It was ~160MiB before and is ~371MiB now.
> 
> Update the test function _get_disk_vm() to get the raw XFS size, because that's the one which should be identical to the block device size.

out of curiosity - what exactly is the log you are refering to ?

kofrezo:
> > When building VMs we require a specific disk size and that ends up being the block device size. When a filesystem is created in this block device it might have some overhead. This overhead was quite small on older Debian releases but on Trixie it has grown quite a bit. The growth is mostly attributed to the log. It was ~160MiB before and is ~371MiB now.
> > Update the test function _get_disk_vm() to get the raw XFS size, because that's the one which should be identical to the block device size.
> 
> out of curiosity - what exactly is the log you are refering to ?

Clarified internally.

